PR #25, opened 2021-05-31
Adding Data to the Databank

larabort:


ohsOllila (changes requested):
I noticed that in the entry for this data set 
http://doi.org/10.5281/zenodo.209362 
has
FF_SOURCE: CHARMM-GUI v1.7 
However, only initial structure has been taken from CHARMM-GUI in here. The force field parameters are from from MacKerell lab website (http://mackerell.umaryland.edu/charmm_ff.shtml). Therefore, this field should be changed to 
FF_SOURCE: MacKerell lab website (http://mackerell.umaryland.edu/charmm_ff.shtml)

It might be also a good idea to check that other entries do not have the same issue.

larabort:
Yes, you are right. In those three data sets there was this mistake. I corrected them.

PR #29, opened 2021-06-22
Adding new molecules

larabort:


larabort:
Here on GitHub it looks like not all entries are indented the same way. But when I open the document on my local computer, everything is on one line.

I hope it is okay like this.

larabort:
Now after I pushed the one for DLPC it is correct again. So now it should work.

PR #59, opened 2021-09-22
Adding new Data

larabort:
The second commit has a wrong name. There are all the files I necessarily had to commit so that I can commit new data.


larabort:
By the way, I do not exactly know where the OrderParameter files come from in the second commit. (I did not run any script except of the AddData.py, as we have discussed.) I think they are still there due to my previous runs of the QualityEvaluation script.

ohsOllila:
There are no new order parameter files committed by you. However, order parameters files have been appeared to Git because I have run the order parameter calculation code myself and committed the results in the Git. The code finds new entries that you have put and calculates the order parameters. So everything goes now as planned.

PR #101, opened 2022-04-19
Merge main to development to test order parameter calculation with new mapping file notation

ohsOllila:


ohsOllila:
I will merge manually

PR #125, opened 2022-11-23
Lipid names are now taken from lipids+dict.

pbuslaev:
I have changed the handling of lipid names, this should solve some issues with united atom force fields. Also, I moved all the checks to the `Parser` to keep main loop simpler


pbuslaev:
This relates to #2

PR #126, opened 2022-11-25
Added extrapolation mode for time estimation

pbuslaev:


pbuslaev:
Solves the issue for #2

PR #137, opened 2023-07-20
adding first alkane sim

awurl:
I'm adding the first lipid/alkane simulation (DPPC/C20). More to follow.


ohsOllila (changes requested):
Please remove the *pyc files at  Scripts/BuildDatabank folder from the commit. These files should not be included to keep the main branch clean.

ohsOllila (approved):

PR #143, opened 2023-09-05
200 ns OpenFF Quality Analysis of POPC

PEFrankel:


ohsOllila:
I am merging this to development branch now for testing.

PEFrankel:
Is the README file the only necessary addition for the repository to collect data or should I commit data using the scripts(I didn't get order parameters after adding the sim)? I assumed there is a wrapping script or a git action that uses the properties.py script after submission of the README. Let me know if I need to commit more or if my submission as is will be viewable in the GUI.

ohsOllila:
Thanks for the data. 

Yes, the README is sufficient (actually the info.yaml file is the minimum). I am in process of writing clear instructions for further steps that could be performed either by a contributor or by someone else. I hope to get this information online in this week. While doing this, I will run the process also for your system and commit the data in the databank. It might take a while before this will be visible in GUI, but if you are interested on how this compares with experiments, there are also other tools to make such plots and quality evaluations. I am also including usage of these tools in the documentation that I am writing.

ohsOllila:
The first version of the documentation is now online: https://nmrlipids.github.io/

I used this system as an example in the script that plots the basic properties of  membranes: 
https://github.com/NMRLipids/databank-template/blob/main/scripts/plotSimulation.ipynb
You can see the comparison with experiments from there.

It is now included also in the ranking lists created by another example notebook: 
https://github.com/NMRLipids/databank-template/blob/main/scripts/plotQuality.ipynb

PEFrankel:
Awesome! Now, if only the sim was any good...

ohsOllila:
There seems to room for for improvement in the sn-2 chain. Improving this might also improve the form factor. 

I would be interested to hear more on how these parameters have been created. Are these generated from the open force field intiative? Is there some documentation available yet how these have been created? If these are created without any explicit or implicit bias to get order parameters or area per lipid correctly, the result is actually pretty good. All the current lipid parameters have either explicit or implicit bias to get these parameters correctly.

PEFrankel:
Yes, it's interesting that the overall quality is decent while the form factor is so far off(and the APL being so low). I suspect charge discrepancies may be one cause for the low quality. It was also submitted without a prior control test of an already ranked force field so my methods may be at fault. I'm testing that now.

The parameters were generated via the [toolkit](https://docs.openforcefield.org/projects/toolkit/en/stable/) and [interchange](https://docs.openforcefield.org/projects/interchange/en/stable/) conversion(early development) from the OpenFF initiative(plenty more documentation online). This submission was a default test of the OpenFF small-molecule parameterization methods without explicit bias for lipids. What "current lipid parameters" are you referring to when mentioning parameter bias? OpenFF, or a different set of parameters?

ohsOllila:
I would maybe first check if there is something strange in dihedral potentials of sn-2 chain. It is possible that fixing inaccuracies in sn-2 would also improve the form factor.

With "current parameters" I refer to all other available lipid parameters that I am aware of, and that are in the databank. My understanding is that none of these force fields would predict reasonable membrane properties without explicit bias for lipids. Nevertheless, if this is the case for OpenFF, it looks really promising to me. Can you parametrize also other than PC lipids?

PEFrankel:
Thanks for the tip! Also, we can parametrize more lipids. Hence the small-molecule versatility. I'll test more lipid compositions soon.

ohsOllila:
We are preparing the revised version of the NMRlipids databank manuscript. Because we now have also a simulation with OpenFF force field, we should add this into table S6 with the citation, see here https://github.com/NMRLipids/DataBankManuscriptText/blob/main/SI.pdf. Is there some citation that we could use here for the OpenFF parameters?

PEFrankel:
There was an error in the previous submission with an inaccurate dOH dipole and missing exclusions in my TIP3P topology (I mentioned running the controls). I just recently clarified cutoff vs PME vdw schemes with this sim and there's more to check considering OpenFF parametrization is one-of-a-kind in this databank. I will get the correct water sim peer-reviewed, then resubmit to the databank soon. When do you plan on finalizing the manuscript?

ohsOllila:
I hope to get the revised version of the manuscript submitted within a week or so. I mean table S6 listing the force force fields with the references:

image

ohsOllila:
@PEFrankel can you send me your email address to my email?

PR #176, opened 2024-04-02
Get rid of one-line JSONs in OP-quality files

comcon1:
Dear maintaners! Here is small but important improvement about JSON dumping. I added compact JSON encoder to store JSON files not in one line. Here I changed only format of storing OP-quality files:
```
{
    "10.1039/c2cp42738a": {
        "M_C3_M M_C3H3_M": [-0.420139, 0.0109258, 0.000684203, -0.394515, 0.02, 0.0337643],
        "M_C4_M M_C4H41_M": [-0.365131, 0.0179467, 0.00112386, -0.351865, 0.02, 0.936608],
        "M_C4_M M_C4H42_M": [-0.23537, 0.0325027, 0.0020354],
        "M_C6_M M_C6H6_M": [-0.068707, 0.0411209, 0.00257509],
        "M_C7_M M_C7H71_M": [-0.305163, 0.0257912, 0.00161511, -0.394515, 0.02, 0.0027106],
...
```

Just two commits:
1. I added file with JSON encoder (source is in the header) and used it in quality-recalculation script (now that's only in one place)
2. Apply this format-change to all Databank.

It is very good practice if we store these files in GIT not to write them into one line. Because now it's much easier to watch into DIFFs after quality recomputing. In GIT-GUIs I see imidiately what's happening after recomputing quality if JSON are stored in that way.  Also, as DIFFs becomes smaller, synchronization time of the project should not growth that dramatically with every JSON-changing commit.


ohsOllila:
Thanks, this is good. It might be a good idea to apply this also to other than quality evaluation json files at some point.

ohsOllila:
This would be particularly useful for ranking files made by makeRanking.py code.

PR #182, opened 2024-04-17
GUI Update compatibility

fsuarezleston:
This pull request is based on https://github.com/NMRLipids/Databank/pull/180 

`databankLibrary.py`

The dictionaries with the definitions were moved into an external file, `databank_defs.py`. This reduces the number of packages needed in the server of the GUI to process the data, and it should not affect the previous code.

`NMRLipids_Update.py`

The script used to update the DB behind the GUI. It can be used to update the whole DB or just for a selection of systems. Keys to access are not included for safety.

`createPDBs.py`

File was modified because it didn't work. However it will not be used for now on.


ohsOllila:
It seems that this pull request would cancel some changes made in the previous one where some case sensitivites were removed. This may be partially due to slow acceptance of pull request, sorry about this. Hopefully you can fix this easily?

PR #202, opened 2024-11-12
Traffic monitoring

korbinib:
Suggested workflow for traffic monitoring.

Either setup a "traffic" branch or change branch in traffic.yml


ohsOllila:
@comcon1 please check that this does not interfere with your developments.

comcon1 (commented):
This CI/CD action add commits to the separate branch on a regular base. It doesn't interfere with any code changes of the code-base. 

But before merging it, please think about you are sure that you want to store traffic information as  commits of THIS repository. I would offer to store at separate repository. We have for example nmrlipids.github.io repository which is updated on a regular base as far as the documentation is updated. I read the documentation of `sangonzal/repository-traffic-action` which Korbenian uses and it's written that YAML can be easily set up to fetch statistics from other repository.

PR #210, opened 2025-02-06
Dev

JHoeflich1:
Add info files 810/info.yaml 811/info.yaml


comcon1:
I will approve after I make sure that README is generated correctly from this INFO.

JHoeflich1:
Sounds good. It looks like some things need to be fixed before I try to merge anyways... I wonder if I can ask it here:

I am having a problem with calculating Order parameters. When running python calcOrderParameters.py, the progress bar completes/gets to 100%, but then I get the error : 

Calculation failed for a7f/9f3/a7f9f36feaa77791483dadb0157f86a6a519f2c3/a30f01e8771c44dc59b78753d1dd27ca67920e38/
Object of type float32 is not JSON serializable.
 
Looks like the file POPCOrderParameters.dat has OP data inside but the POPCOrderParameters.json file is empty. 
Any idea where this error could be coming from?

comcon1:
Tell me please your OS and python / MDAnalysis versions? Can you also show us .dat file created if it is non-empty.

-------- Оригинальное Сообщение --------
06.02.2025 23:52, JHoeflich1  написал:

JHoeflich1:
Yes here is the info requested:

OS/ versions: Ubuntu 22.04.4 LTS, python 3.10.0, mdanalysis 2.8.0

POPCOrderParameters.dat in txt file:
[POPCOrderParameters.txt](https://github.com/user-attachments/files/18712106/POPCOrderParameters.txt)

[outerr.log](https://github.com/user-attachments/files/18712107/outerr.log)
[condalist.txt](https://github.com/user-attachments/files/18712108/condalist.txt)

comcon1:
Thank you very much Julianne, I'm suspecting that we never actually tested on numpy-2.x. I will now try to create the naked environment and run tests. This error is easy to fix, but I'm afraid it can be not the only error.

comcon1 (approved):
During the discussion new issue was created, and I made PR #212. I have checked that analysis works well  with INFO-files provided, so it can be merged.

Issue #211, opened 2025-02-10
Analysis scripts fail


comcon1:
> I am having a problem with calculating Order parameters. When running python calcOrderParameters.py, the progress bar completes/gets to 100%, but then I get the error : 
>
```
Calculation failed for a7f/9f3/a7f9f36feaa77791483dadb0157f86a6a519f2c3/a30f01e8771c44dc59b78753d1dd27ca67920e38/
Object of type float32 is not JSON serializable.
```
 
>Looks like the file POPCOrderParameters.dat has OP data inside but the POPCOrderParameters.json file is empty. 
Any idea where this error could be coming from?

_Originally posted by @JHoeflich1 in https://github.com/NMRLipids/Databank/issues/210#issuecomment-2641300795_

comcon1:
> OS/ versions: Ubuntu 22.04.4 LTS, python 3.10.0, mdanalysis 2.8.0

> POPCOrderParameters.dat in txt file:
> [POPCOrderParameters.txt](https://github.com/user-attachments/files/18712106/POPCOrderParameters.txt)

> [outerr.log](https://github.com/user-attachments/files/18712107/outerr.log)
> [condalist.txt](https://github.com/user-attachments/files/18712108/condalist.txt)

_Originally posted by @JHoeflich1 in https://github.com/NMRLipids/Databank/issues/210#issuecomment-2644067215_

PR #212, opened 2025-02-10
Fixing some bugs arising when I run tests with numpy-2.x

comcon1:
As it was identified in PR https://github.com/NMRLipids/Databank/pull/210, and then isolated as a bug #211, we have some problems on numpy 2. One small problem has a small fix at 19334a53c811725d79fcdf71de1fb692b0ab9821

Also I fixed tests in 69b5b118e1e4d436f0b00ffb4878d13d943f5a55 which were broken after I updated cardiolipin mapping files.

Finally, we have some problems with numerically unstable algorithms. These problems I don't solve but I improve precision consideration in comparison of float numbers. By commit a07641373203f87e6243e9540c934601f0f56ed5 it is important how many significant digits are saved, and not just an absolute error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ The code has been tested in Linux environment with python 3.7 or newer and recen
 Setup using conda as distribution:
 
 ```
- $ conda create --name databank python==3.10 MDAnalysis periodictable -c conda-forge
+ $ conda create --name databank python==3.10 'numpy<2.0' MDAnalysis periodictable -c conda-forge
  $ conda activate databank
  $ (databank) conda install tqdm yaml -c conda-forge
 ```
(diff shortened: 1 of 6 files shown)


batukav:
I agree with the sentiment in commit a076413 

@comcon1 Do you have a python environment that runs all the tests successfully? If so, we should export that env details and provide it as an `environment.yaml` file.

batukav:
> I agree with the sentiment in commit [a076413](https://github.com/NMRLipids/Databank/commit/a07641373203f87e6243e9540c934601f0f56ed5)
> 
> @comcon1 Do you have a python environment that runs all the tests successfully? If so, we should export that env details and provide it as an `environment.yaml` file.

comcon1:
I just avoided `numpy>2` explicitly in `requirements.txt` until we are not ready with this numerical instability stuff. Exporting whole environment I would consider redundant. There are so many explicit version information in it. `numpy<2` works well, I tried these form-factor tests now again.

PR #218, opened 2025-02-17
Continue PR #212 after branch rebasing

comcon1:
I had to re-create PR #212 because some other PR was accepted and I wanted to rebase the branch.
@batukav I did approximately what you asked, could you please review, because I want to move forward.

### DIFF
--- a/Scripts/tests/test_analyze.py
+++ b/Scripts/tests/test_analyze.py
@@ -94,7 +94,7 @@ def systemLoadTraj(systems):
 # ----------------------------------------------------------------
 # Every test function is parametrized with system ID to make clear reporting
 # about which system actually fails on a test function.
-
+MAXRELERR_COMPARE_THRESHOLD = 2e-6
 
 def compareJSONsBtwSD(jsfn: str):
     """Function to call from the test to compare JSON to the precomputed one.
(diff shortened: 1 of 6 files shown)


ohsOllila:
This is waiting review from @batukav, right?

Scripts/tests/test_analyze.py
```diff
@@ -122,15 +122,19 @@ def compareJSONsBtwSD(jsfn: str):
         assert len(j1) == len(j2), \
             f"Problem in {jsfn} comparison: lists has different lengths!"
         for k1 in range(len(j1)):
-            _a = (np.array(j1[k1]) - np.array(j2[k1])).ravel()
-            assert (_a**2).sum() < 1e-7, (
+            _a = np.abs(np.array(j1[k1]) - np.array(j2[k1])).ravel()
+            _b = np.abs((np.array(j1[k1]) + np.array(j2[k1])).ravel())/2
+            _b[_b == 0] = MAXRELERR_COMPARE_THRESHOLD
+            assert np.max(_a/_b) < MAXRELERR_COMPARE_THRESHOLD, (
```
batukav, line 128:
```
for k1 in range(len(j1)):
            np.testing.assert_allclose(
                np.array(j1[k1]),
                np.array(j2[k1]),
                rtol=MAXRELERR_COMPARE_THRESHOLD,
                err_msg=(
                    f"Problem in {jsfn} comparison: line {k1}.\n"
                    f"Computed: {j1[k1]} \n"
                    f"Precomputed: {j2[k1]}"
                )
)
```

seems like a safer and more elegant approach. with the current proposal, if `j1[k1] = -j2[k1]` , and `j1[k1] < E-12` the test will succeed but I suppose it shouldn't, as signs seem to have been predicted incorrectly. 

if you want to keep the current solution, then I'd suggest setting `_b[_b==0]` to a different but very small value.

batukav (changes requested):

Scripts/tests/test_analyze.py
comcon1: Thanx. No, I don't want to use bicycles. As you showed me how to avoid a bicycle, I am using your solution: 8f730bf

comcon1:
I tested rebasing of this branch on my fork: there were no conflicts, and it passed all the tests. So I'm ready for merge+rebase.

batukav (commented):
It all looks good to me, we can merge this whenever you want

batukav (approved):

PR #219, opened 2025-02-21
Adding patch to monolayer branch for-non-databank-paper-version

comcon1:
1. Adding electron density blurring to cover better match exp vs MD XRR data
2. Clear the code
3. Add possibility to analyse united-atom force fields
4. Add some simulations including DOPS at different protonation state


ohsOllila (approved):
I assume that explanation of the structure of this part of the databank is coming (or is it already somewhere)?

comcon1:
> I assume that explanation of the structure of this part of the databank is coming (or is it already somewhere)?

We have a notebook in the other repository (with notebooks) which explains, how to use the functions here. It should be anyway rewritten completely within new NMRlipids structure in main branch, so it's not a production-level code. It is just a code that works. We need it to be delivered even to a separate branch because we are submitting the paper which cites the functions implemented. Mostly `calcXRR` function.

PR #220, opened 2025-04-08
fix path of GUI-update scripts as it works in a truncated environment…

comcon1:
As we discovered, the automatic site update is broken. There is a fix. Also, I added some instructions to build up the minimal environment required for running the code.


ohsOllila (approved):
This can be merged.

Issue #217, opened 2025-02-16
Normalize documentation generation


comcon1:
1. Documentation should be removed from `Scripts/BuildDatabank` forder.
2. I don't see now how it's generated. It's completely unknown for me. I don't see anything here:
https://github.com/NMRLipids/Databank/tree/main/.github/workflows
And I don't have rights to see "deployments". I know how to generate sphinx build, but I don't see what is the process behind its delivery to the site.
3. I don't know actually if the documentation synchronization works properly together with @korbinib statistics, which he put there.

ohsOllila:
This is a bit messy indeed. If I remember correctly, I have done it such that I manually copy the content of this folder: 
https://github.com/NMRLipids/Databank/tree/main/Scripts/BuildDatabank/docs/build/html
into this repo:
https://github.com/NMRLipids/NMRLipids.github.io

This is because I had to get it done somehow. I do not know what would be the correct way to do this.

comcon1:
> This is a bit messy indeed. If I remember correctly, I have done it such that I manually copy the content of this folder:

Ok. So you don't have continuous integration at this part? Then we should definitely fix it, yes.

ohsOllila:
Yes, there is no continuous integration. This would good to fix with relatively high priority because up-to-date documentation is very important, especially now when updates are being implemented. Do you have an idea how to implement this?

comcon1:
Seems to be done. Will close after checking how it works.

PR #228, opened 2025-05-19
Add workflow to build and deploy documentation 

MagnusSletten:
This PR brings in automatic building and deployment of documentation with changes within Scripts folder. 

Once PR's are merged to main with relevant changes, the documentation will be built and then deployed [here](https://github.com/nmrlipids/NMRLipids.github.io) which in turn will update Github Pages deployment. 

The github actions happen within a docker container from an image on [dockerhub](https://hub.docker.com/repository/docker/nmrlipids/doc-builder/general)

The purpose behind using a docker container here is simplifying handling of dependencies and creating a consistent running environment. Even though this is just documentation building, Sphinx is still required to recognize the imports. In this case a docker image has been created with installed dependencies from the requirements file within the project and uploaded as a docker image on dockerhub.  There is a step for updating the dependencies within the workflow in case there are new ones, but this will be quick since the rest is already installed.

I will add more documentation on docker and a small tool to help maintain docker images in a separate PR.

It also makes sense to make an organization on Dockerhub so that images can be handled in a better way as a team. 

Example of previously deployed documentation, just with different deployment [location](https://magnuspriv.github.io/NMRLipids.github.io/)

The actual build folder (where built html files are) is deleted and the documentation-building folder is moved out of BuildDatabank.


comcon1:
This PR fixes #217

comcon1:
> The github actions happen within a docker container from an image on dockerhub
Who does Docker Hub belong to now? I am not allowed to go to your link and see the containers. Is it assumed to be closed? You assumed organization for Dockerhub - I think, it's a proper way, can you start it, and then we take over?

> Even though this is just documentation building, Sphinx is still required to recognize the imports. 
Does Sphinx need to recognize imports for developmental parts (e.g., unit tests) as well? In my new PR, I've separated `requirements.txt` and `requirements-dev.txt`. Is the dev-file the file that should be imported by Sphinx?

MagnusSletten:
> Who does Docker Hub belong to now? I am not allowed to go to your link and see the containers. Is it assumed to be closed? You assumed organization for Dockerhub - I think, it's a proper way, can you start it, and then we take over?

I was thinking about this some more this morning and @korbinib also had some ideas regarding the use of docker. Maybe we should make an issue on this separately.  A docker organisation would be very practical for teamwork, but I need to check their pricing model properly. If I understand it correctly I could convert my user-account to an organization without incurring costs, but it is somewhat limited. One alternative improvement would be passing this account on to one of the admins here then I (and others) can be added as a contributors for the relevant repositories @comcon1 

> Does Sphinx need to recognize imports for developmental parts (e.g., unit tests) as well? In my new PR, I've separated `requirements.txt` and `requirements-dev.txt`. Is the dev-file the file that should be imported by Sphinx?

As long as there are python files which is part of the documentation created which has specific imports, those imports needs to be recognised. Maybe the additional dev imports aren't used in the files that is part of the documentation through doc-strings right now, but they may be so later. We can just switch to requirements-dev for the environment, there's no real downside.

korbinib:
One possible alternative as discussed would be to use the [GH container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) but the storage is limited to 500MB on the free plan.

korbinib (approved):

comcon1 (approved):
I approve.

I checked triggering - it's reasonable. I checked how the folders are moved - also OK. 

probably we need to mention in comments where Dockerfile from `nmrlipids/doc-builder` will live. But that we can do after we organize this repo for containers.

PR #250, opened 2025-06-06
Mediumclean history branch 

### LINKED ISSUE #249: Create Data-cleaned version of the main branch
Data-clean version of the main branch should be prepared and tested before being changed to main.

comcon1:
This PR must close #249 issue. Everything should work in this branch as it is supposed to work in the main in future.


MagnusSletten:
I've ran tests within this branch without any unexpected behaviour and tested my workflows related to tests and doc-building, all without any errors. 

Running the submodule commands per the README instructions also works fine with the Data from the main data repository. 

Inspecting the git diff does not reveal any unexpected changes to files.

Scripts/BuildDatabank/check_mappings.py
```diff
@@ -0,0 +1,70 @@
...
+              debugging mappings code.
+"""
+
+import os
+import sys
+import MDAnalysis as mda
+from tqdm import tqdm
+import traceback
+
+from DatabankLib import NMLDB_SIMU_PATH, lipids_set
```
MagnusSletten, line 15: This import of lipids_set does not work, but its error is not directly related to this PR. 

``` from DatabankLib.settings.molecules import lipids_set ``` 

The above works.

MagnusSletten (commented):
I still cannot see any major problems with these changes. I've done some more testing, the only issue I had is unrelated to this PR directly.

PR #267, opened 2025-06-18
Use MAICoS for profile & form factor calculations

PicoCentauri:
This replaces the old form factor code to use the MAICoS analysis library instead.

The code will currently not work because we need a relase of MAICoS to include the electron density code...

# TODO

- [x] Use the code inside `analyze.py`
- [x] Check testsuite

### DIFF
--- a/Scripts/DatabankLib/requirements-dev.txt
+++ b/Scripts/DatabankLib/requirements-dev.txt
@@ -11,3 +11,4 @@ scipy
 flake8
 pep8-naming
 pytest
+pytest-check


comcon1:
@PicoCentauri, how are we going to work on it? Will we open the development branch for it and then collaborate there? Merging to main after that?

PicoCentauri:
I think we can just work in this branch, right?

I think the first thing is to make the `computeMAICOS` function actually work by preprocessing the trajectories and passing them to the initialization of the universe.

comcon1:
I see. It's a draft PR from maicos branch of our main repo. So we just commit to it from both sides. Currently, I must spend time to prepare my application. But I will try to continue processing this at least with some speed..

comcon1:
You can try this now by yourself
```
pytest -vs --cmdopt sim1 test_analyze.py::test_analyze_maicos
```
It returns me the following:
```
Calculation failed for aa4/
ProfilePlanarBase.__init__() missing 1 required positional argument: 'sym_odd'
Traceback (most recent call last):
  File "/home/comcon1/repo/Databank/Scripts/DatabankLib/analyze.py", line 1023, in computeMAICOS
    form_factor = FormFactorPlanar(
  File "/home/comcon1/repo/Databank/Scripts/DatabankLib/maicos.py", line 44, in __init__
    super().__init__(
TypeError: ProfilePlanarBase.__init__() missing 1 required positional argument: 'sym_odd'
```

Probably we need to specify a specific MAICoS branch in requirements.txt?

PicoCentauri:
Arghh we had a recent bugfix

https://github.com/maicos-devel/maicos/pull/497

I am going to do a maicos release today, fix the error here and we can afterwards pin the version in the requirements

Sorry for hustle!!!

PicoCentauri:
Yeah it is good that you kicking. Otherwise I will forget everything I did xD

comcon1:
Now it works. What I run for now. You can run the same command on your machine:
``` 
cd Scripts/tests/
export NMLDB_TEST_NOWIPE=1
pytest -vs --cmdopt sim1 test_analyze.py::test_analyze_maicos[566-1]
```
Then we can compare `Scripts/tests/Data/Simulation.1/aa2/WaterDensity_mcs.json`:
```
[
    [-32.1956, 0.323456, 0.00200582],
    [-31.8961, 0.322393, 0.00209888],
    [-31.5966, 0.323343, 0.00194509],
    [-31.2971, 0.32897, 0.0020889],
    [-30.9976, 0.326125, 0.0020778],
    [-30.6981, 0.329095, 0.00194097],
...
```
 and `Scripts/tests/Data/Simulation.2/aa2/WaterDensity.json`:
```
[
    [-3.13266, 325.131],
    [-3.09804, 321.851],
    [-3.06343, 327.616],
    [-3.02881, 323.17],
    [-2.9942, 325.116],
    [-2.95958, 325.541],
...
```
1. MAICoS Z axis in in Angs. Legacy code returns nm. 
2. Electron density is given in units "x1e3".

comcon1:
Form factor against precomputed looks good. But what are the units? I multiplied one by 970 to get the other
image

PicoCentauri:
> Form factor against precomputed looks good. But what are the units? I multiplied one by 970 to get the other

The units should be number of electrons per Å^3. What units do you want?

comcon1:
> This is the correct error if the sample are uncorrelated which probably is not the the case here.

OK. It actually sent me some messages about how it wasn't able to estimate correlation times. You mentioned in the presentation that MAICoS corrects according to these correlation times by default, right?

> The units should be number of electrons per Å^3. What units do you want?

That's electron density. Our are x1000, so it's electrons per nm^3. So, you have (Ang, e/Ang^3) and we need (nm, e/nm^3).

The form factor units are not clear to me. Do you know the multiplier there? It seems that if I multiply it by 100, it looks almost the same. Can it be 100?

PicoCentauri:
> OK. It actually sent me some messages about how it wasn't able to estimate correlation times. You mentioned in the presentation that MAICoS corrects according to these correlation times by default, right?

Yes, if there are not enough samples we can't estimate the correlation time.

> That's electron density. Our are x1000, so it's electrons per nm^3. So, you have (Ang, e/Ang^3) and we need (nm, e/nm^3).

Sounds reasonable. If you convert can you add a comment on the conversion?

> The form factor units are not clear to me. Do you know the multiplier there? It seems that if I multiply it by 100, it looks almost the same. Can it be 100?

Yes I think I also had 100 when I tested. There was some hardcoded factor in your form factor script without explanations...

comcon1:
OK. Units we fixed. You can double-check if I did right.

comcon1:
@PicoCentauri look, I tried to introduce purified legacy element recognizer here https://github.com/comcon1/Databank/tree/maicos

But I don't know if it is correct now to override with guesser. Because I define elements via atom.element = 'NA' and you expect that element will be some object. That I haven't yet understood. Need more time.

PicoCentauri:
Unit changes look good.

Regarding the element guesser: Yes, as I already wrote I it is very likely that this might be not ideal. A guesser from your side would be ideal.


>But I don't know if it is correct now to override with guesser. Because I define elements via atom.element = 'NA' and you expect that element will be some object. That I haven't yet understood. Need more time.

This looks good:

```python
elements.guess_elements(system, u)
u.guess_TopologyAttrs(force_guess=["elements"])
```

you can remove the last line as you already overwrite the elements. Elements are no objects. They are just simple characters.

comcon1:
@PicoCentauri I've done some stuff:
- I rebased **maicos** branch to the current HEAD because some stuff has already been done, and some conflicts already appeared
- I implemented united-atom 'elements'. Now it complains on the MAICoS side:

```
  File "/home/comcon1/.conda/envs/databankdev/lib/python3.10/site-packages/maicos/lib/weights.py", line 69, in density_weights
    raise KeyError(
KeyError: "Element 'Ch2' not found. Known elements are listed in the `maicos.lib.tables.elements` set."
```

The minimal test you can run to probe this behavior is:
```
pytest -vs test_analyze.py::test_analyze_maicos[243-1]
```

PicoCentauri:
Thanls for the work!

Hmm united atom types should be supported. Can you maybe try CH2. I am on holidays but maybe have some time to check if there is a missing logic in maicos.

comcon1:
> Hmm united atom types should be supported. Can you maybe try CH2. I am on holidays but maybe have some time to check if there is a missing logic in maicos.

I actually use 'CH2' because that is what we use in . So it's not me who does the second letter lower-case. It's probably MDAnalysis's internal polishers.

comcon1:
> In fact I forgot the united atom element support. Will be fixed with
> 
> [maicos-devel/maicos#510](https://github.com/maicos-devel/maicos/pull/510)
> 
> and we do a release once it is merged.

Thanx that you write! Seems that it actually works. Will complete my tests so

Scripts/DatabankLib/requirements-dev.txt
```diff
@@ -1,6 +1,7 @@
 buildh>=1.6.1
 MDAnalysis>=2.7.0
-numpy>=1.20.0,<2.0
+numpy>=2.0
+maicos>=0.11,<0.12
```
PicoCentauri, line 4: Good! I think we should also bump the dependency to at least 0.11.2.

comcon1:
image

comcon1:
We are ready for review! Note that MAICoS specific branch is required to test the PR!

Scripts/DatabankLib/analyze.py
```diff
@@ -550,59 +557,86 @@ def computeOP(  # noqa: N802 (API)
...
+        "Diporder2Water.json",
+        "TotalMassDensity.json",
+        "TotalDensity.json",
+        "WaterMassDensity.json",
+        "LipidDensity.json",
+        "WaterDensity.json",
+        "LipidChargeDensity.json",
+        "DielectricLipid",
+        "DielectricWater",
+        "DielectricTotal"
```
PicoCentauri, line 588: Is there a reason why these files are no `.json`?

Scripts/DatabankLib/analyze.py
```diff
@@ -625,171 +659,326 @@ def computeFF(  # noqa: N802 (API)
...
+            u = mda.Universe(struc_name, trj_name)
+
+        # -- PHILIP code starts here --
+        # We us a hardoced bin width
+        bin_width = 0.3
+
+        # Guesser might charged elements wrong like Cl- may guessed as Cl...
+        from DatabankLib.settings import elements
+        u.guess_TopologyAttrs(force_guess=["elements"])
+        elements.guess_elements(system, u)
```
PicoCentauri, line 788: I think we can remove the MDA guesser since our own guesser does the same job.

```suggestion
        from DatabankLib.settings import elements
        elements.guess_elements(system, u)
```

Scripts/DatabankLib/maicos.py
```diff
@@ -0,0 +1,164 @@
...
+    def encode(self, o):
+        if isinstance(o, np.ndarray):
+            return CompactJSONEncoder.encode(self, o.tolist())
+        else:
+            return CompactJSONEncoder.encode(self, o)
+
+
+# TODO: Maybe perform unit conversions.
+# Units in maicos/MDAnalysis are Å for length, e for charge, u for mass
+# See https://mdanalysis.readthedocs.io/en/latest/documentation_pages/units.html
```
PicoCentauri, line 24: I think we did the unit conversion.

```suggestion
```

PicoCentauri (commented):
Thanks for the nice work @comcon1. The guesser looks solid!

PicoCentauri:
> @PicoCentauri, please, convert this PR to non-draft once you merge your uff-electron branch

Yes the branch is merged and inside the 0.11.2 release.

comcon1:
Ok. I will. I am fixing numbers for regression tests now.

-------- Оригинальное Сообщение --------
17.07.2025 12:18, Philip Loche  пишет:

MagnusSletten:
> Ok. I will. I am fixing numbers for regression tests now.

I am getting a few errors from test_api, this might be less relevant since you are fixing the numbers currently:

```
FAILED Scripts/tests/test_api.py::test_GetFormFactorMin[281-result0] - assert np.float64(3.857005800000041e-05) < 1e-09
FAILED Scripts/tests/test_api.py::test_GetFormFactorMin[566-result1] - assert np.float64(6.428603399999916e-05) < 1e-09
FAILED Scripts/tests/test_api.py::test_GetFormFactorMin[243-result2] - assert np.float64(4.667207999999751e-06) < 1e-09
FAILED Scripts/tests/test_api.py::test_GetFormFactorMin[86-result3] - assert np.float64(5.993338599999956e-05) < 1e-09
FAILED Scripts/tests/test_api.py::test_analyze_th[281-1-4.18335] - assert 0.016610000000000014 < 1e-05
FAILED Scripts/tests/test_api.py::test_analyze_th[243-1-4.27262] - assert 0.013149999999999551 < 1e-05
```
and 2 related to computeMAICOS. This might be more relevant for debugging: 

```
Traceback (most recent call last):
  File "/github/workspace/Databank/Scripts/DatabankLib/analyze.py", line 786, in computeMAICOS
    elements.guess_elements(system, u)
  File "/github/workspace/Databank/Scripts/DatabankLib/settings/elements.py", line 93, in guess_elements
    selection.atoms.elements = selection.n_atoms*[elname]
    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/MDAnalysis/core/groups.py", line 321, in __setattr__
    raise AttributeError(
AttributeError: Cannot set arbitrary attributes to a Group
```

PicoCentauri:
Ohh the latter might be indeed relevant. Let me check if this is because we removed some function calls this morning.

comcon1 (approved):
Now, we are ready to go. I found that numbers become again slightly different for MAICOS, but when I build plots, they are absolutely the same. Also, I tried to use different versions of numpy/scipy, and the numbers didn't further change. So, it looks numerically stable now.

PicoCentauri:
Looks very good. Thanks for the update of the logic behind!

Issue #268, opened 2025-06-18
Documentation generate empty pages


comcon1:
Almost all auto-gen files becomes empty by some reason..
https://nmrlipids.github.io/auto_gen/Scripts.BuildDatabank.makeRanking.html

PR #270, opened 2025-06-18
Fix bug with building documentation after data separation 

MagnusSletten:
Specific imports failed without having data submodule present which caused automated files not being produced.

Related to #268 

This has been tested in my fork.

### DIFF
--- a/.github/workflows/BuildDocumentation.yml
+++ b/.github/workflows/BuildDocumentation.yml
@@ -20,6 +20,7 @@ jobs:
             uses: actions/checkout@v4
             with:
                ref: "main"
+               submodules: True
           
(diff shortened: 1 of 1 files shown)


comcon1 (changes requested):
I see. It need to be able to import. This solution does work, but we do not need Data for building documentation. It's enough in principle to just plug-in test data like Markus did for templates.

This is what I did for example to make documentation buildable locally
```
export NMLDB_DATA_PATH=~/repo/Databank/Scripts/tests/Data
export NMLDB_SIMU_PATH=~/repo/Databank/Scripts/tests/Data/Simulations.1/
```

it will run everything work without any data.

comcon1 (approved):
Bra

Issue #275, opened 2025-06-23
AddData.py adds extra lines to README.yml causing errors


MagnusSletten:
**Describe the bug**
Running AddData.py on any info.yml file produces README.yml with an unintended line at the top for instance: 

```
 !!python/object:DatabankLib.core.System
_store:
  DOI: 10.5281/zenodo.11614468
  SOFTWARE: gromacs
  TRJ:
  - - 566.xtc
  TPR:
```
The first line should not be added. 

I suspect there is just a one line change needed in AddData.py. Maybe:

```
with open(outfile_dict, "w") as f:
yaml.dump(sim, f, sort_keys=False, allow_unicode=True)
```

should be:

```
 with open(outfile_dict, "w") as f:
 yaml.dump(sim.readme, f, sort_keys=False, allow_unicode=True)
```

comcon1:
Yes, the fix is this but more important that it is not caught by tests.

PR #276, opened 2025-06-23
Fix add data problems

comcon1:
- Test that AddData construct readable `README.yaml` which can be read by `initialize_databank()`
- Fix behavior of test files
- Fix bug #275

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -11,14 +11,16 @@ else
   exit 1
 fi
 
+status=0
 # Run the tests using pytest
 if command -v pytest &> /dev/null; then
-  pytest Scripts/tests --cmdopt sim2
-  pytest Scripts/tests --cmdopt sim1
-  pytest Scripts/tests --cmdopt adddata
-  pytest Scripts/tests --cmdopt nodata
+  pytest Scripts/tests --cmdopt sim2 || status=1
+  pytest Scripts/tests --cmdopt sim1 || status=1
+  pytest Scripts/tests --cmdopt adddata || status=1
+  pytest Scripts/tests --cmdopt nodata || status=1
 else
   echo "pytest is not installed. Please install required libraries using"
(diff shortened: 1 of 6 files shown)


runtests.sh
```diff
@@ -15,6 +15,7 @@ fi
 if command -v pytest &> /dev/null; then
   pytest Scripts/tests --cmdopt sim2
   pytest Scripts/tests --cmdopt sim1
+  pytest Scripts/tests --cmdopt adddata
   pytest Scripts/tests --cmdopt nodata
```
MagnusSletten, line 19: I noticed that there's an issue with error codes when running the tests in this way (mostly relevant for CI), essentially if the last test has no error then the error code will be zero even if previous tests fail. 

I fixed it locally by doing something like:

```
status=0
# Run the tests using pytest
if command -v pytest &> /dev/null; then
  pytest Scripts/tests --cmdopt sim2 || status=1
  pytest Scripts/tests --cmdopt sim1 || status=1
  pytest Scripts/tests --cmdopt adddata || status=1
  pytest Scripts/tests --cmdopt nodata || status=1
else
  echo "pytest is not installed. Please install required libraries using"
  echo ">> pip install -e . -r Scripts/DatabankLib/requirements-dev.txt <<"
  status=1
fi
exit $status
``` 
So if one individual test returns non zero error code then status becomes 1 and is returned as exit code.

MagnusSletten (commented):
I've ran tests locally and they work fine. Introduction of new AddData test is very good and the registration of the test markers reduces the output when running tests which is good.

The runtests behaviour is something I noticed earlier so I can include the changes there in my PR on tests today or we can update it here.

comcon1:
@MagnusSletten, I invited you to my repo. Could you please commit your recommendations directly to this PR?

PR #281, opened 2025-06-26
fix docker documentation not instructing mounting the drive correctly

batukav:
Fixing a typo in the docker documentation: the instructions incorrectly mounted the current working directory to the WORKDIR of the docker container.

### DIFF
--- a/Scripts/Docker/README.md
+++ b/Scripts/Docker/README.md
@@ -46,12 +46,12 @@ Go to the directory where you have the Databank repository. Then,
 
 1. Start the container with your code mounted:
    ```bash
-   docker run -it -v $(pwd):/workspace NAME_OF_THE_DOCKER_IMAGE
+   docker run -it -v $(pwd):/github/workspace NAME_OF_THE_DOCKER_IMAGE
    ```
    or
    if you pulled the latest image,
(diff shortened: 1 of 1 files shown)


MagnusSletten:
> Btw, the tests did run very quickly for me (<5 minutes). Did we change something regarding how the tests run? @MagnusSletten @comcon1

Yes, the more intensive regression tests are in a separate file now: `runtests-regression.sh`

Those are run on a weekly basis but not when pull requests are made.

MagnusSletten (approved):
Good catch with seeing this error! Tested locally and it works fine with your changes.

batukav:
cool, thanks for the info.

MagnusSletten:
There is also some more changes to the readme i'll add in another PR soon. Right now we still have

> Inside container, create a new virtual environment using Conda as previously described. (Must update this)

The image got changed quite a bit during the PR process because I figured out how to avoid some of the annoying permission problems I had with github standard runners. So I could combine the two images into one. Because at that point the only difference would be python venv vs Conda.

But if you see utility in building more on a separate dev container (for instance with uv), maybe "nmrlipids/dev-container" you can definitely do that.

PR #284, opened 2025-07-03
Seperate info file processing from global methods on simulations

MagnusSletten:
Separating scripts that make global changes from those that only update files for new simulations.

This is done to reduces the chance of merge conflicts when multiple simulations are added to the data repository via website prior to them getting merged.


comcon1 (approved):
That's OK. Let's start with this. Further will check how to plug-in IDs

Issue #263, opened 2025-06-16
Speed-up OP claculations


comcon1:
Trying somehow..

batukav:
I'll try this out.

PR #286, opened 2025-07-08
Order parameter code refactor and speed up

batukav:
This pull request deals mainly with the refactoring of the order parameter code. Other small changes include:

1- Allowing `runtests-regression.sh` to accept `pytest` arguments from the CLI
2- Adding and improving some logging functionality for the `test_analyze.py` code. Those changes can be extended to the other test scripts, if wanted
3- Simple code formatting

We knew for some time that the order parameter code is rather slow and recently decided to refactor it to increase its performance #263 . Within this pull request, I mainly adopted a vectorized implementation for the order parameters. Instead of reading the trajectory frame by frame and calculating the order parameters per frame one by one, we calculate the order parameter for all residues simultaneously at each frame.

I ran the test suite a few times. According to the output, we are getting around 20x speed-up with the new implementation while reproducing the results of the original implementation (as far as the tests tell). For instance, if one runs `pytest` with the flag `-vvl --durations=0`, one will get something like the below timings (left original -> right new implementation)

```
404.30s call     test_analyze.py::test_analyze_op[281-1] -> 19.47s call     test_analyze.py::test_analyze_op[281-1]
```
As order parameter computation is a very very central and important part of the databank, I would feel much more comfortable if multiple people can reproduce these results before merging. 

@comcon1 , not sure if I could find any examples, but how can I run the order parameter code on a given INFO file? I would like to test the code with multiple trajectories that are not part of the automated tests.

@MagnusSletten, I requested your review as well, the more the merrier for this PR.

### DIFF
--- a/Scripts/DatabankLib/databankop.py
+++ b/Scripts/DatabankLib/databankop.py
@@ -1,5 +1,4 @@
 """
-    :meta private:
  calculation of order parameters of lipid bilayers
  from a MD trajectory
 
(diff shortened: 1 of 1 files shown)


comcon1 (approved):
OK, let's merge then.

Issue #291, opened 2025-07-18
Recorded problems with AddData


comcon1:
1. [x] Incorrect determination of the center-of-bilayer
2. [x] SHA1 is too slow
3. [x] first we should check mappings and then download+sha1

comcon1:
I think, we need to close this one and open separated issues instead..

PR #292, opened 2025-07-19
Addfile improve manythings

### LINKED ISSUE #187: Improve AddFile.py robustness
`AddFile.py` shoud first check if `gmx` command exist in the shell and then start doing something. Otherwise, it spend time for downloading and then crashes.

### LINKED ISSUE #285: remove "work_dir" key from the README.yaml
work_dir key should not be present from the final README.yaml files as this 1- has nothing to offer for the reproducibility of the computation, 2- reveals personal information.

copy of https://github.com/ohsOllila/IDPdatabank/issues/14#issuecomment-3031634387

comcon1:
This PR mainly deal with #291. It introduces `--dry-run` behavior of `AddData.py` where only first 50MB is loaded. It should be always enough to allow script finish working. After dry-run is done, one can run a real run.
Also closes #285.
Also partly fixes #187.

### DIFF
--- a/Scripts/BuildDatabank/AddData.py
+++ b/Scripts/BuildDatabank/AddData.py
@@ -44,13 +44,13 @@ from MDAnalysis import Universe
 import DatabankLib
 from DatabankLib.core import System
 from DatabankLib.databankLibrary import (
-    calc_file_sha1_hash,
-    create_databank_directories,
     lipids_set,
     molecules_set
 )
 # helpers
 from DatabankLib.databankio import (
+    calc_file_sha1_hash,
+    create_databank_directories,
     download_resource_from_uri,
     resolve_download_file_url
 )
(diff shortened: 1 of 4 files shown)


Scripts/BuildDatabank/AddData.py
```diff
@@ -88,15 +88,20 @@
     )
     parser.add_argument(
         "-w", "--work-dir",
-        help="set custom temporary working directory [not set = read from YAML]",
-        default=""
+        help="set custom temporary working directory [not set = /tmp]",
```
MagnusSletten, line 91: Maybe we can at this point just remove the requirement of work_dir being present in info file within method: parse_valid_config_settings

Within this method there is:

```
   # STEP 3 - check working directory
    if "DIR_WRK" not in sim:
        raise KeyError("'DIR_WRK' Parameter missing in yaml")
```
Then I can just remove this entry on the form entirely.

comcon1:
@MagnusSletten, this PR will give me the possibility to debug an INFO-file before running your automatic runner. 

Also, it removes WRK_DIR, so you can also remove it from the interface.

PR #296, opened 2025-07-22
Update methods related to simulation processing after Databank changes. 

MagnusSletten:
* For simulation processing a dry run of AddData.py is used prior to the real run to quickly discover potential errors. 
* Some code is also changed to account for changes related to analysis method changes. 
* It will use logger instead of regular print statements

### DIFF
--- a/Scripts/WorkflowScripts/ProcessInfoFile.py
+++ b/Scripts/WorkflowScripts/ProcessInfoFile.py
@@ -17,8 +17,12 @@ def main(info_file_path):
     base_tmp = os.path.join(parent_folder, "databank_workdir")
     work_directory_dry  = os.path.join(base_tmp, "dry")
     work_directory_real = os.path.join(base_tmp, "real")
-    os.makedirs(work_directory_dry, exist_ok=True)
-    os.makedirs(work_directory_real, exist_ok=True)
+    try:
+        os.makedirs(work_directory_dry, exist_ok=True)
+        os.makedirs(work_directory_real, exist_ok=True)
+    except OSError as e:
+        logger.error(f"Failed to create work directories")
+        sys.exit(1)
 
     run_python_script(
         path_dict["AddData_path"],
--- a/Scripts/WorkflowScripts/Workflow_utils.py
+++ b/Scripts/WorkflowScripts/Workflow_utils.py
@@ -2,7 +2,6 @@ import subprocess
 import sys
 import os
 import logging
-
 """
 Contains methods used for python scripts related to workflows.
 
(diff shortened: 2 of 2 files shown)


Scripts/WorkflowScripts/Workflow_utils.py
```diff
@@ -38,13 +46,14 @@ def run_python_script(script_path, args=None, error_message="Python script faile
     if args is None:
         args = []
     try:
+        logger.info(f"Running python script with path {script_path}")
         subprocess.run(
             [sys.executable, script_path, *args],
             check=True,
             cwd=working_dir  
         )
     except subprocess.CalledProcessError:
```
batukav, line 55: Why only capturing this exception only? To my knowledge, other exceptions are also possible (https://docs.python.org/3/library/subprocess.html#exceptions) and thus should be captured as well.

Scripts/WorkflowScripts/ProcessInfoFile.py
```diff
@@ -10,20 +10,30 @@
...
     path_dict = get_databank_paths(NMLDB_ROOT_PATH)
-    work_directory = "/tmp/databank_workdir"
-    os.makedirs(work_directory, exist_ok=True)
+    parent_folder = os.path.dirname(NMLDB_ROOT_PATH)
+
+    base_tmp = os.path.join(parent_folder, "databank_workdir")
+    work_directory_dry  = os.path.join(base_tmp, "dry")
+    work_directory_real = os.path.join(base_tmp, "real")
+    os.makedirs(work_directory_dry, exist_ok=True)
+    os.makedirs(work_directory_real, exist_ok=True)
```
batukav, line 21: could you wrap these with try/except ?

Scripts/WorkflowScripts/Workflow_utils.py
```diff
@@ -10,6 +11,13 @@
    repository can safely ignore it.
 """
 
+logging.basicConfig(
+    format="%(asctime)s [%(levelname)s]: %(message)s",
+    datefmt="%I:%M:%S %p",
```
batukav, line 16: since `logger` is being used by many other scripts, maybe we should consider standardizing its usage throughout the codebase.

Scripts/WorkflowScripts/Workflow_utils.py
MagnusSletten: Now I'm capturing OSError, ValueError and the parent of CalledProcessError which is SubProcessError. 

Anything more? I generally want any sort of exception to cause direct errors in CI since it should be obvious that something has failed within github actions. Catching many types of errors and immediately exiting for a general helper method to run scripts is normally not good, but it's fine for this specific purpose.

I could move the immediate exit handling to the main method itself though, then the helper methods would work better generally speaking since exceptions can be handled in less strict ways if it's desired.

Scripts/WorkflowScripts/ProcessInfoFile.py
MagnusSletten: Did this just now 👍

Scripts/WorkflowScripts/Workflow_utils.py
MagnusSletten: Having it standardized would be really nice. Any thoughts on this specific use of the logger or do we leave this for later?

PR #309, opened 2025-08-16
Add workflow to build docker images monthly 

### LINKED ISSUE #307: Docker building automation
CI/CD features ensure that docker image is rebuilt every monthly.

MagnusSletten:
Workflow related PR which includes:

* Adds workflow that builds and pushes each of the docker images currently in use. 
* Adds dockerfile for the doc-builder image that is currently in use
* Fixes small permission problem in core image

The builds without dependencies are run in parallel. The jobs for building gromacs and the core image can be combined if we want a bit less code duplication at the cost of slightly messier logs. 

Edit: I added a commit combining the jobs.

Closes #307

### DIFF
--- a/.github/workflows/BuildDockerImages.yml
+++ b/.github/workflows/BuildDockerImages.yml
@@ -9,26 +9,13 @@ jobs:
   build_gromacs:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout GROMACS repo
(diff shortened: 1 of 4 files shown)


comcon1:
I agree. Rebuilding docker-images on non-tagged commits is too expensive. We already have quite a lot of free github CI/CD.

Issue #312, opened 2025-08-16
MAICOS: Unexpected token water - element attribution error


comcon1:
```
06:17:09 PM [INFO]: attribute elements has been guessed successfully.
06:17:13 PM [ERROR]: Calculation failed for fc9/adc/fc9adc57cdc847019621267bfed6cc71d05d93a7/cddacbf14fa088499729ab009228a3d6f25b38ea/
06:17:13 PM [ERROR]: Selection failed: 'Unexpected token 'water''
06:17:13 PM [ERROR]: Traceback (most recent call last):
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/DatabankLib/analyze.py", line 840, in computeMAICOS
    elements.guess_elements(system, u)
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/DatabankLib/settings/elements.py", line 75, in guess_elements
    selection = u.select_atoms(selstr)
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/MDAnalysis/core/universe.py", line 759, in select_atoms
    return self.atoms.select_atoms(*args, **kwargs)
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/MDAnalysis/core/groups.py", line 3553, in select_atoms
    selections = tuple(
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/MDAnalysis/core/groups.py", line 3555, in <genexpr>
    selection.Parser.parse(
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/MDAnalysis/core/selection.py", line 1575, in parse
    parsetree = self.parse_expression(0)
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/MDAnalysis/core/selection.py", line 1583, in parse_expression
    exp1 = self._parse_subexp()
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/MDAnalysis/core/selection.py", line 1607, in _parse_subexp
    raise SelectionError(errmsg) from None
MDAnalysis.exceptions.SelectionError: Selection failed: 'Unexpected token 'water''
```

hsantila:
yes, aka the one made issue for in BilayerData: https://github.com/NMRLipids/BilayerData/issues/83 with some suggested solutions and a separate one at the MDAnalysis github

PR #313, opened 2025-08-17
Allow MAICoS to work with dummy atom

### LINKED ISSUE #311: MAICoS: Element 'D' not found.
```
03:28:15 AM [INFO]: Analysing 52061 trajectory frames.
03:28:15 AM [DEBUG]: Module input: DiporderPlanar(*args).run(*args)
03:28:16 AM [ERROR]: Calculation failed for 4f7/da7/4f7da716a581be4f931d2e9a2890872bec19c66a/4f7da716a581be4f931d2e9a2890872bec19c66a/
03:28:16 AM [ERROR]: "Element 'D' not found. Known elements are listed in the `maicos.lib.tables.elements` set."
03:28:16 AM [ERROR]: Traceback (most recent call last):
  File "/opt/conda/envs/conda_container_env/lib/python3.10/site-packages/maicos/lib/weights.py", line 34, in _resolve_electron_count
    return electron_count[element.title()]
KeyError: 'D'
```

comcon1:
MAICoS should process dummy atoms properly. If it has a charge, it should probably process the charge to compute the dipole moment. But if we compute electron density, a dummy element should be introduced. I think, it could be 'X'. It could be also empty string ''. But 'X' is safer because '' is unclear if it is just unset or it's dummy.

I wrote small unit-test which tests maicos integration on a dummy on-the-fly-created trajectory. 
@PicoCentauri, let's modify PR to pass these tests.

I mention @hsantila here because I'm actually not understanding what happened in the previous version with dummy atoms. We have at least two mapping files with dummies, and these dummies are named in a completely different way: D1, D2 in one and Vi in the other. I don't see any signs of code parts that deal somehow with these dummy atoms. I would consider all the form-factor data for all trajectories related to these water molecules to be corrupted and **required to be immediately deleted** from the BilayerData. 

In the end, it will close #311 and #312, probably too.

### DIFF
--- a/Scripts/DatabankLib/analyze.py
+++ b/Scripts/DatabankLib/analyze.py
@@ -37,6 +37,7 @@ from DatabankLib.databankio import resolve_download_file_url
 from DatabankLib.databankop import find_OP
 from DatabankLib import analyze_nmrpca as nmrpca
 from DatabankLib.maicos import (
+    is_system_suitable_4_maicos,
     FormFactorPlanar,
     DielectricPlanar,
     DensityPlanar,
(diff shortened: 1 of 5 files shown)


comcon1:
> In the museum databank, it seems the electrons were guessed by mapping file name. There was an explicit exception written in for "D":

Ok. Exception for D I see. But for "Vi" I don't. And TIP4P and OPC4 are marked M_Vi_M. I have PR in BilayerData when I change universal name to X*. M_X_M M_X1_M ... and then we will pass 'Dummy' to maicos by this new guesser. Please review it then https://github.com/NMRLipids/BilayerData/pull/56

Yes, that can of worms I don't want to touch. I guess that for some realization of drudes we can even modify the orientation of dipolar moment by just drude particles.

Issue #319, opened 2025-08-19
Restart calculation if zenodo fails to download


comcon1:
This error led to skipping the whole process for one property. 
```
09:28:04 AM [INFO]: Computing APL (Area Per Lipid) for all systems
Warning: AM [WARNING]: netCDF4 is not available. Writing AMBER ncdf files will be slow.
09:28:35 AM [INFO]: Filtering systems by range: (None, 0)
09:28:35 AM [INFO]: Number of systems in databank: 1
09:28:35 AM [INFO]: System title: IMM_TLCL_divided_PA_PG_310K
09:28:35 AM [INFO]: System path: 785/ea1/785ea16d65161e6ae5d5bf3018f6c3f2540b4826/8227880b80c102d686b502adeecbbf73e0768b74/
Error: 6 AM [ERROR]: Calculation failed for 785/ea1/785ea16d65161e6ae5d5bf3018f6c3f2540b4826/8227880b80c102d686b502adeecbbf73e0768b74/
Error: 6 AM [ERROR]: HTTP Error 502: Bad Gateway
Error: 6 AM [ERROR]: Traceback (most recent call last):
  File "/__w/BilayerData/BilayerData/Databank/Scripts/DatabankLib/analyze.py", line 177, in computeAPL
    u = system2MDanalysisUniverse(system)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/__w/BilayerData/BilayerData/Databank/Scripts/DatabankLib/databankLibrary.py", line 317, in system2MDanalysisUniverse
    trj_url = resolve_download_file_url(doi, trj)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/__w/BilayerData/BilayerData/Databank/Scripts/DatabankLib/databankio.py", line 178, in resolve_download_file_url
    raise hte
  File "/__w/BilayerData/BilayerData/Databank/Scripts/DatabankLib/databankio.py", line 163, in resolve_download_file_url
    _ = urllib.request.urlopen(uri, timeout=10)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/urllib/request.py", line 216, in urlopen
    return opener.open(url, data, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/urllib/request.py", line 525, in open
[...]

PR #320, opened 2025-08-20
Retry with exponential backoff

batukav:
closes issue #319 

All the http requests are wrapped with the "retry_with_exponential_backoff" decorator. It will try to complete the request 5 times with increasing intervals before failing. This should prevent many transient errors from the server. Note that for 4XX errors the code quits without retrying

### DIFF
--- a/Scripts/DatabankLib/databankio.py
+++ b/Scripts/DatabankLib/databankio.py
@@ -19,66 +19,138 @@ from typing import Mapping
 import urllib.error
 from tqdm import tqdm
 import urllib.request
+import functools
 
 logger = logging.getLogger(__name__)
 MAX_DRYRUN_SIZE = 50 * 1024 * 1024  # 50 MB, max size for dry-run download
 
+def retry_with_exponential_backoff(max_attempts=3, delay_seconds=1):
+    """
+    Decorator that retries a function with exponential backoff for transient network errors.
+    """
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            attempts = 0
+            current_delay = delay_seconds
+            while attempts < max_attempts:
+                try:
+                    return func(*args, **kwargs)
+                
+                # --- New logic to handle non-retriable HTTP client errors ---
+                except urllib.error.HTTPError as e:
+                    # Check if the error is a client error (4xx) which is not likely to be resolved by a retry.
+                    if 400 <= e.code < 500:
+                        logger.error(
+                            f"Function {func.__name__} failed with non-retriable client error {e.code}: {e.reason}"
+                        )
+                        raise e  # Re-raise the HTTPError immediately
+                    
+                    # For other errors (like 5xx server errors), proceed with retry logic.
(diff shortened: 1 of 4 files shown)


Scripts/DatabankLib/databankio.py
```diff
@@ -19,66 +19,138 @@
 import urllib.error
 from tqdm import tqdm
 import urllib.request
+import functools
 
 logger = logging.getLogger(__name__)
 MAX_DRYRUN_SIZE = 50 * 1024 * 1024  # 50 MB, max size for dry-run download
 
+def retry_with_exponential_backoff(max_attempts=3, delay_seconds=1):
```
comcon1, line 27: That looks cool.

Scripts/DatabankLib/databankio.py
```diff
@@ -97,101 +169,96 @@ def update_retrieve(self, b=1, bsize=1, tsize=None):
...
-            except TimeoutError:
-                raise RuntimeError(f"Cannot open {uri}. Timeout error.")
+                # Use the decorated helper to check if the URI exists
+                with _open_url_with_retry(uri):
+                    pass
             except urllib.error.HTTPError as hte:
+                # Special handling for 429 after retries have failed
                 if hte.code == 429:
-                    if sleep429/5 > 10:
+                    if sleep429 / 5 > 10:
```
comcon1, line 234: Why not to implement these tryouts using the decorator you introduced? It will be much nicer.

Scripts/tests/test_loads.py
```diff
@@ -72,6 +72,75 @@ def test_goodDOI__resolve_doi_url(self):
...
                 dio.resolve_doi_url('10.5281/zenodo.8435138', True))
 
+    @staticmethod
+    def _create_mock_success_response(mocker):
+        mock_response = mocker.MagicMock()
+        mock_response.read.return_value = b"mocked content"
+        mock_response.getheader.return_value = "100" # Content-Length
+        mock_response.code = 200
+        mock_response.reason = "OK"
+        return mock_response
```
comcon1, line 82: that looks very good

Scripts/tests/test_loads.py
```diff
@@ -72,6 +72,75 @@ def test_goodDOI__resolve_doi_url(self):
...
+            ),
+            (
+                "persistent URLError fails",
+                lambda m: URLError('persistent error'),
+                ConnectionError,
+                5,
+            ),
+            (
+                "non-retriable 403 HTTPError fails immediately",
+                lambda m: HTTPError('url', 403, 'Forbidden', {}, None),
```
comcon1, line 101: this way is also very nice

comcon1 (changes requested):
The code looks supercool. The only question is why not to implement 429-tries with your decorator.

comcon1 (approved):
OK. THat's good. I also was watching about how are your changes behave in production. And it seems OK. But now I realized that everything was predownloaded. But if it will be some error, I will open another issue. Now we can go.

PR #321, opened 2025-08-23
Add json schema validation for info files

### LINKED ISSUE #290: YAML verification by schema
> @MagnusSletten Why have only order parameters and NMRPCA been computed? Is it possible to view logs from the runner?

From the logs I see one obvious problem that is quick to fix here. There will also be more files added after the merge when the "global" methods are run, but parts of the analysis is broken due to some input being interpreted as a string when it should be integer/float. I'll fix this. 

Ideally the info-file check would raise errors prior to this point though, I'll take a look at it as well.

_Originally posted by @MagnusSletten in https://github.com/NMRLipids/BilayerData/issues/28#issuecomment-3075350145_

MagnusSletten:
Adds json schema for validating info files:
- Checks required keys, types, number of required entries
- Software dependent requirement for certain keys like gromacs -> tpr
- Unknown keys are allowed in the main info files due to point number [Number 19 here](https://nmrlipids.github.io/schemas/simulation_metadata.html)

Wrote some quick sanity tests to see that the validation worked as expected. I used the pytestmark "adddata" because it will be used in that context and the only data needed is info-files, maybe there's a better fit or we create a new test-mark. I also add the repo root to path prior to one import to make it work.

JSON schema and validation methods are located within BuildDatabank/ SchemaValidation, let me know if there's a more fitting place for it.  

I did not add exception handling for the base methods as I thought it might be better to let callers handle these, but if it's the preference to do it here, then I'll add it in quickly. 

Running this validation against all info.yml files in the data repository revealed the [following the errors](https://gist.github.com/MagnusSletten/d86b988ad81f81437377be325435a1fa)

Closes #290

### DIFF
--- a/Scripts/DatabankLib/requirements-dev.txt
+++ b/Scripts/DatabankLib/requirements-dev.txt
@@ -12,3 +12,4 @@ flake8
 pep8-naming
 pytest
 pytest-check
+jsonschema 
--- a/Scripts/tests/test_yaml_validation.py
+++ b/Scripts/tests/test_yaml_validation.py
@@ -1,0 +1,110 @@
+import pytest 
+import os
+import sys
+import copy 
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__),'../', '../')) 
+
+from BuildDatabank.SchemaValidation.ValidateYAML import validate_info_file, validate_info_dict
+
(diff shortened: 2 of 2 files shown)


batukav (approved):
Looks pretty good to me. Well done!

Issue #329, opened 2025-08-29
Error when trying to import module


mdondrup:
**Describe the bug**
Trying to import the Databank module after installation with pip yields a nonsensical error message:

```
python GUI_DB_Update.py -c config.json -s ../../BilayerData
Traceback (most recent call last):
  File "/Users/me/github/BilayerGUI_laravel/Python/GUI_DB_Update.py", line 26, in <module>
    import DatabankLib
  File "/Users/me/micromamba/lib/python3.13/site-packages/DatabankLib/__init__.py", line 37, in <module>
    raise RuntimeError(
    ...<2 lines>...
            "to git-clonned `Databank` folder.")
RuntimeError: Seems that you installed package in a non-debug mode. In this case you *must* set NMLDB_ROOT_PATH explicitly to git-clonned (sic!) `Databank` folder.
```
This may be a relic from code separation:

When setting the environment:` export NMLDB_ROOT_PATH=$HOME/github/Databank`

```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/me/github/Databank/Data/Molecules/membrane'
```

When setting `export NMLDB_ROOT_PATH=$HOME/github/BilayerData` (where the Molecules are now):
 - getting the original run-time error.

**To Reproduce**
 - Check out the Databank repo (main branch)
 - `pip install ./Databank` (works)
 -  

```
import DatabankLib
import DatabankLib.core

```

 
5. Specify python and numpy versions.
Not relvant...
**Expected behavior**
Modules can be imported without error message or error message reads:

RuntimeError: Seems that you installed the package in non-debug mode (explain what that means). In this case you *must* set/export  ** the environment variable ** NMLDB_ROOT_PATH explicitly to point to the folder containing the BilayerData.

- Code import works if conditions are met

**Actual behavior**
RuntimeError....

comcon1:
If you want to have separated BilayerData repository, just specify DATA env, and that's it. You don't need to specify ROOT at all.

mdondrup:
`If you want to have separated BilayerData repository, just specify DATA env, and that's it. You don't need to specify ROOT at all.`

That does not work with the original code either.

comcon1:
Without ROOT but with DATA:
```
(databankprod) comcon1@tanteliss:~/repo/test$ export NMLDB_DATA_PATH=/home/comcon1/repo/test/BilayerData
[...]

PR #330, opened 2025-08-29
Fixed issue #329 Error when trying to import module

mdondrup:
- Now allows to import the module if NMLDB_DATA_PATH is set
- Fixed the error message, typo and clarity
- Set some default and added the link to the Data repository

### DIFF
--- a/Scripts/DatabankLib/__init__.py
+++ b/Scripts/DatabankLib/__init__.py
@@ -38,7 +38,7 @@ NMLDB_DATA_PATH: str = os.environ.get(
 
 # If the environment variable is not set, we assume that the data folder is in the root path
 if NMLDB_DATA_PATH is None:
-    NMLDB_DATA_PATH = os.path.join(NMLDB_ROOT_PATH, 'BilayerData')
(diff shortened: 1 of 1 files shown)


comcon1:
> 1. The checks are the same as the were.
> 2. You can check if ROOT path was actually set before you check Data. If ROOT is set but Data is not clonned, then the message should be different.
> 3. You probably would like to change `README.md` regarding these paths' description.  Because there it is also written about debug and non-debug mode installing and you don't like it.

It's not a very strong opinion, we can discuss here how we present. But if you start fixing documentation messages about it, let's do it in a final way in this PR.

mdondrup:
I have converted the PR to draft, it is possibly best to start it in its own branch. I see that a few more changes to documentation and messages are required, but my priority was to find out how to get the module importable for moving the GUI_update script out of the Datanbank into the BilayerGUI repository. After I have finished the GUI update script, I am open to work on some more refactoring.

comcon1:
This PR is not relevant anymore after introducing new behavior in #343

Issue #326, opened 2025-08-27
Perform memory check for PCA relaxation times


comcon1:
> So essentially there is a [threshold](https://github.com/NMRLipids/Databank/blob/91c877d8ec00f6030107a0400751152ab52c0a40/Scripts/DatabankLib/analyze_nmrpca.py#L68) for memory in the current version.
> 
> What is happening with larger trajectories, is that a [short trajectory is created](https://github.com/NMRLipids/Databank/blob/91c877d8ec00f6030107a0400751152ab52c0a40/Scripts/DatabankLib/analyze_nmrpca.py#L223).
> 
> What can be modified:
> - smaller memory cutoff can be used (e.g. `1.5Gb`)
> - the skipping step for short trajectory can be estimated as `int(1.5 * traj_size / memory_cutoff)`
> 
> If this sounds like a plan, it is a 2 line modification 

 _Originally posted by @pbuslaev in [#316](https://github.com/NMRLipids/Databank/issues/316#issuecomment-3224840142)_

PR #331, opened 2025-09-01
solves #326

pbuslaev:
- Added flexible definition of skipping frequency. For large trajectories, skipping frequency is defined as 1.1 traj.size / memory_cutoff
- Memory cutoff can be set as global variable `MEM_CUTOFF`

### DIFF
--- a/Scripts/DatabankLib/analyze_nmrpca.py
+++ b/Scripts/DatabankLib/analyze_nmrpca.py
@@ -6,20 +6,19 @@ PCAlipids analysis.
 
 For details check:
 
-[Principal Component Analysis of Lipid Molecule Conformational
-Changes in Molecular Dynamics Simulations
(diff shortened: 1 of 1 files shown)


pbuslaev:
I think it is worth documenting `MEM_CUTOFF` global var. I'll check where should it go in docs

comcon1 (approved):
I added the description of the enviorment variable. 

I'm not sure that I'm happy with the solution, by the following reasons. 
1. I don't see why we save all the trajectory while making short if we could only save lipids. And even just heavy atoms of lipids.
2. I don't see the point of comparing the size of whole trajectory if it's actually single-lipid who is eating size. If it's a single-lipid system, it will eat much more from memory than it is 20-lipid system 10-times more in size.

However, this will work better than before, and we can live with it, and now we need to close it because Batu is waiting.

Further improvements into the PCA modules will probably happen afterwards as far as we started to work on it.

PR #332, opened 2025-09-03
Introducing Ruff

batukav:
With this pull request we are introducing `ruff`, a Python linter and code formatter, to the project. 

I already updated many of the files with 

`ruff check --fix`

There are around 2500 more errors to be fixed. My proposal is to fix these errors slowly file by file and finish by the time we publish the repo as the pip package. I can handle all the changes requested by `ruff`.


batukav:
@MagnusSletten It is possible to add the `pre-commit` hooks using `ruff` to the `github-actions`. You have more experience there than me, could you please take care of that implementation?

comcon1:
1. I changed the required docstring convention. We currently use sphinx/RTD. Not numpy or google. 
2. We should remove pep8-naming as ruff has naming convention inside
3. If ruff needs dependencies, we should update `requirements-dev.txt` and not `requirements.txt`

Let's go through at least some files and see how to adjust the complaints. We currently have old-style paths, so we need to remove many PTH until we decide to switch to `Path` system. Also, there is other stuff which I don't like.
@batukav @MagnusSletten, please check which rules I currently changed.

@MagnusSletten, can you try to satisfy for some of your files? Just to get an idea of which rules we want to ignore.

comcon1:
I did some of the things which should have been done, I suppose. I will wait for your opinion about the current **ruff** complaints. Remember that we shouldn't accept side PRs while we are here. It's an important step. We should do it carefully.

MagnusSletten:
I went through the workflowfile folder and fixed errors ruff pointed out, most of them are now gone except for errors linked to the workflowfile folder itself using namespace imports. To avoid this same error happening with schemavalidation I also moved those to DatabankLib in one commit, if we want to consider other options here I can quickly drop this commit. 

```
102 Missing `from __future__ import annotations`, but uses PEP 604 union
  --> Scripts\WorkflowScripts\Workflow_utils.py:23:30
   |
22 | def run_python_script(
23 |      script_path: str, args: list|None  = None,
   |                              ^^^^^^^^^
24 |      error_message: str="Python script failed", working_dir: str|None = None) ->None:
25 |     """
   |
help: Add `from __future__ import annotations`
```

Having working_dir: str|None with python 3.9 gives complaints because this type notation was introduced in 3.10, but there are other simple ways to satisfy this requirement. Any downside to bumping the requirement from 3.9 to 3.10?

comcon1:
Achtung. I tried with 3.9 environment, it doesn't work anyway because maicos requires 310 min. So we cannot stick to 39 anyway, and I will have to increase requirements by 0.1

comcon1 (approved):
I'm ready to go

comcon1:
@batukav I was thinking what to do but you have already merged back main to ruff. What is the next step?

batukav:
I just resolved the conflict between the incoming changes and the existing code in `Scripts/DatabankLib/analyze_nmrpca.py` 

If you have no other objections, we can merge the PR and move on.

comcon1:
> If you have no other objections, we can merge the PR and move on.

I made one post-merge commit to this file and want to run regression tests again.

comcon1:
Ready now. Tests ok.

Issue #325, opened 2025-08-27
Modify compute_databank command line arguments for computing only FF and TotalDensity


comcon1:
Let's add the key `--ff` for form-factor&density and leave `--maicos` for the whole set of MAICoS files.

Issue #323, opened 2025-08-26
Reorganize documentation menu


ohsOllila:
Documentation is currently in some kind of development version: https://nmrlipids.github.io/index.html

We should keep a reasonable version there before we are ready to put a properly updated version. 

There are a lot of information missing now, including the universal atom names etc. This is the place where I go to check them and many other things myself when I need to check something. I cannot work efficiently if the documentation is not accessible. Therefore, we should keep it in some kind of reasonable state even during development.

ohsOllila:
Ok, I can find these files now. However, the documentation is still a bit messy right now. I would like to have it in some kind of reasonable format all times. Are you able to fix it soon?

For example, CSC people are asking more information and it is difficult to provide because documentation is not clear now.

MagnusSletten:
> Ok, I can find these files now. However, the documentation is still a bit messy right now. I would like to have it in some kind of reasonable format all times. Are you able to fix it soon?

Yes, after #308 is resolved I can make a small PR with documentation changes. 

Some improvements:
* Links to the actual mapping files are outdated due to separation. 
* Make universal molecule and atom name documentation easier to find

Are there other problems you see with the documentation? Now it's organised into three main categories: Data Organization, Data Contribution and Databank Standalone Programs.

ohsOllila:
Ok, now it looks better.

However, at least example applications part refering to codes in databank-template repo is missing. At least this should be added. Because I do not see the original documentation anymore, I do not know if something else is also missing. Would it be possible to see somehow the old version to compare?

MagnusSletten:
> However, at least example applications part refering to codes in databank-template repo is missing. At least this should be added. Because I do not see the original documentation anymore, I do not know if something else is also missing. Would it be possible to see somehow the old version to compare?

I haven't yet added the links but here is an older version(from the beginning of June) of the documentation if you want to compare:  https://magnuspriv.github.io/NMRLipids.github.io/

ohsOllila:
For me the main issue is that links to relevant content in the column on the left hand side are not there anymore, i.e., this in the old documentation:

<img width="276" height="331" alt="Image" src="https://github.com/user-attachments/assets/076bc4b8-f842-48f0-98a9-2018abbaa03f" />

When these are returned, we can close this. Potential improvement of documentation in general is a bigger issue.

comcon1:
I've done it but I don't know how to trigger deployment. @MagnusSletten could you write how to force him to redeploy documentation at nmrlipids.github.io

PR #337, opened 2025-09-10
Some improvements regarding simulation addition

comcon1:
1. compute_databank can now compute only densities+FF. Will close issue #325.
2. Found a severe bug that downloader didn't actually download on filesize mismatch despite this behavior was presumed. Fixed now and fixed a test case as well.
3. Some ruff-related style improvements on databankio and maicos analysis.
4. Small documentation updates regarding system addition covering the current situation.
5. A bit restructured applied to documentation menu to satisfy partly #323.

### DIFF
--- a/Scripts/DatabankLib/maicos.py
+++ b/Scripts/DatabankLib/maicos.py
@@ -1,4 +1,18 @@
+"""
+Wrappers for MAICoS calculations adapted to Databank needs.
+
+- Checks if a system is suitable for maicos calculations
+- Custom JSON encoder for numpy arrays
+- Custom maicos analysis classes with save methods adapted to Databank needs
+"""
+
+import contextlib
 import json
+import os
+import re
+import subprocess
+from collections import deque
+from logging import Logger
 
 import maicos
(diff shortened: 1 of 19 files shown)


Scripts/DatabankLib/maicos.py
```diff
@@ -40,10 +55,180 @@ def is_system_suitable_4_maicos(system: System):
...
+            check=True,
+        )
+    except subprocess.CalledProcessError as e:
+        msg = "Failed during centering on g3 carbons."
+        raise RuntimeError(msg) from e
+
+    try:
+        os.remove(xtcfoo)
+        os.remove(ndxpath)
+    except OSError as e:
```
batukav, line 220: What happens if we get other errors than `OSError`? Would you like to capture them or shall we postpone the error handling to later when I define a consistent error handling scheme?

comcon1:
@batukav, thanks for looking at it. Let's wait for @MagnusSletten also, because I made a small change in the documentation building. I have now checked on my machine here, and it also works.

PR #343, opened 2025-09-16
Restructering the repo

### LINKED ISSUE #200: Restructuring the repo
Hi all,

This issue was brought up by @comcon1 here on Issue #195 and also had been keeping my mind busy for a while. 

Currently we are tracking and versioning both the code (in Scripts directory) and output (Data directory) together. This has the potential to make tracing the history rather difficult in the future, plus could create headache with merge requests. Hence, I'd like to discuss how we can restructure the Databank such that we can version/track the code and data separately. 

One potential solution is to use the Github submodules as @comcon1 suggested. I'm not informed enough of the submodules' capabilities.

Other potential solution is to move the data to a cloud storage and version it using some data versioning tool like `DVC`. This would mean that the data and code will be hosted at different places, might be a bit more difficult to maintain but a neater solution. 

Since we also have the GUI, we should be careful to not break its integration. Unfortunately I'm not well aware of how the GUI integration works, so I also would like to gather some info about it here.

@markussmiettinen , @ohsOllila any suggestions?

PicoCentauri:
Fixes #200

# TODO

- [x] add standalone scripts
- [x] move unit and regression tests into `tox` workflow
- [x] Build documentation based on `tox -e docs`
- [x] Add github workflows for tests, lint etc.
- [x] Build per PR docs based on read the docs
- [x] Add codecov token
- [x] Add developer docs for tox workflows


comcon1:
Dear colleagues. I got a bit stack. I want to distribute zipped test Data folder together with the package, and I managed to do it via tox build (see my commits), but I cannot understand where to put pre-building zip-creation into pyproject.toml to make it working in 'pip install .'

Can you please help?

src/DatabankLib/bin/nml_initialize_data.py
```diff
@@ -0,0 +1,121 @@
...
+
+def download_toy_folder(dest: str = "ToyData", folder_path: str = "Scripts/tests/Data") -> None:
+    """Download a folder from a GitHub repository using the GitHub API."""
+    branch = "main"
+    api_url = f"https://api.github.com/repos/{CODEREPO}/contents/{folder_path}?ref={branch}"
+    os.makedirs(dest, exist_ok=True)
+
+    with urlopen(api_url) as response:
+        if response.status != 200:
+            raise Exception(f"Feil ved hente data: {response.status} {response.text}")
```
MagnusSletten, line 50: Minor change needed here with wrong language used in exception.

comcon1:
I added what I wanted. During my committing, I was inspecting all the changes, so I have already done a kind of review job here on your parts of code. Let's inspect the code added by me then, and we can merge.

src/DatabankLib/bin/initialize_data.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Write environment setup file
+    with open("databank_env.rc", "w") as f:
+        f.write(f"export NMLDB_DATA_PATH={data_path}\n")
+        if sim_path is not None:
+            f.write(f"export NMLDB_SIMU_PATH={sim_path}\n")
+
+    print(f""""Data initialized into {data_path}. Please do
+
+   $ source databank_env.rc
```
MagnusSletten, line 107: This is OS dependent, but maybe we expand to windows in separate PR? Not sure how many besides me use windows on the team. It's not a big change anyways.

MagnusSletten (changes requested):
Looks good! We can make a quick change to the small comments that are the wrong language, maybe we fix the init for windows later.

src/DatabankLib/bin/nml_initialize_data.py
comcon1: This block was removed.

Issue #336, opened 2025-09-09
info.yaml template need to be recreated


comcon1:
It lived before in some development branch 

https://github.com/NMRLipids/Databank/blob/main/Scripts/DatabankLib/SchemaValidation/Schema/info_yml_schema.json

It doesn't exist anymore and I don't know where to find it. Easier to do from scratch. And store together with the schema.

PR #351, opened 2025-10-05
Update docs after PIP-PR (try 3)

comcon1:
I fixed some issues here and there over the documentation. Template page is now appear in the menu as well as metadata of both experiments and simulations' YAML files. Fixed one of the old issues https://github.com/NMRLipids/Databank/issues/336.

----
📚 Documentation preview 📚: https://Databank--351.org.readthedocs.build/en/351/


comcon1:
> I think we can ignore the codecov

I would say we will have to open an issue to satisfy codecov. What he is showing currently is quite meaningful. But this PR doesn't change the coverage according to its report.

MagnusSletten (approved):

PR #352, opened 2025-10-06
Add new experiment metadata

comcon1:
Currently, experimental metadata is added to documentation.

It will close this issue: https://github.com/NMRLipids/BilayerData/issues/97

----
📚 Documentation preview 📚: https://Databank--352.org.readthedocs.build/en/352/

### DIFF
--- a/docs/src/schemas/experiment_metadata.md
+++ b/docs/src/schemas/experiment_metadata.md
@@ -67,11 +67,14 @@ Do not provide whole salts! Only separated ions. Remember that the counterions o
 Dictionary of additional molecules in the format:
 ```
 ADDITIONAL_MOLECULES:
-    TFA: trifluoroacetic acid
-    DMSO: dimethylsulfoxide
-    DMS: trimethylsilylpropanesulfonate
+    TFA: trifluoroacetic acid, 0.1%
+    DMSO: dimethylsulfoxide, 0.1%
+    DSS: sodium trimethylsilylpropanesulfonate, 0.01%
+    EDTA: ethylenediaminetetraacetic acid, 0.1 mM
 ```
-we can use INCHI-key, CAS number or just IUPAC name. If molecule is important for the composition, it will get the metadata inside the databank.
+we can use INCHI-key, CAS number or just IUPAC name. If molecule is important for
+the composition, it should get the metadata inside the databank and be mentioned under
+`SOLUTION_COMPOSITION` instead.
 
 8. **PH**  
 pH of the system (number or UNKNOWN)


docs/src/schemas/experiment_metadata.md
```diff
@@ -1,43 +1,70 @@
...
 
 **Field descriptions**
 
-1. **DOI**  
+1. **ARTICLE_DOI**  
 DOI of the original publication where the experimental data originates.
 
-2. **TEMPERATURE**  
-Temperature of the experiment.
+2. **DATA_DOI**
```
korbinib: doi is nice, but what if the dataset does not have a doi, but only some other PID or even just a URL?

docs/src/schemas/experiment_metadata.md
```diff
@@ -1,43 +1,70 @@
...
+1. **ARTICLE_DOI**  
 DOI of the original publication where the experimental data originates.
 
-2. **TEMPERATURE**  
-Temperature of the experiment.
+2. **DATA_DOI**  
+DOI of the dataset deposition with raw NMR data (e.g., nmrXive).
+
+3. **TEMPERATURE**  
+Temperature of the experiment. If `T_RF_HEATING` is 'no_inromation', the reported temperature from the probe is given. Otherwise, please insert RF-corrected temperature.
```
korbinib: C/K (I assume the latter but this is not explicit)

docs/src/schemas/experiment_metadata.md
```diff
@@ -1,43 +1,70 @@
...
 
-3. **MOLAR_FRACTIONS**  
+4. **T_RF_HEATING**  
+How RF heating is dealt (no_information / measured / guessed)
+
+5. **MEMBRANE_COMPOSITION**  
 Dictionary of molar fractions of bilayer components. For example:
 ```
 MOLAR_FRACTIONS:
```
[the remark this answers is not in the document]
comcon1: Done

mdondrup:
> I don't see the point hardcoding these dictionaries. I see the help in having them as an instructions to a person to fill the fields. Fields can still be just text fields. And I don't see how it affects SQL schema. We need to have a list of additional text fields and different views for different (currently 2 - NMR or FF) types of experiments - that is all we need for now.

Let me simply explain a few things. In software development, we need well-defined tasks and requirements, even or particularly if we are being agile. That was why I was so keen on the active use of a new feature. A few notes to explain what I mean:

1. Your metadata annotations are dictionaries, they are not just text. This is a completely NEW feature request.

```
| MEMBRANE_COMPOSITION | Dictionary of molar fractions of membrane phase |
| SOLUTION_COMPOSITION | Dictionary of ion concentrations in the system |
| ADDITIONAL_MOLECULES| Dictionary of molecules which are not in the databank |
``` 

This is novel functionality. None of the lipid metadata was like that. The current state of the implementation supports only flat 'name-value-type' style of metadata.

2. Dictionaries could be nested, which results in a more complex implementation. SQL isn't really good for modeling 
or traversing graph structures of arbitrary depth or searching through them. It's possible, but it makes the implementation much more complicated. That is why I want to avoid arbitrary-depth nested data if possible.

3. Your example entries imply a certain semantics, but it is not explicit because YAML doesn't support referential semantics. That semantics affects how the implementation needs to be carried out:
   
- They may need to be searchable, e.g. search for an experiment by lipid or molecule. 
- They refer to other entities, e.g. lipids and molecules. Should these entries also be semantically linked, e.g. as foreign key references, should we show a link to the molecule in the interface? (That means we need to parse that entry)
- How should cases be resolved where entries are referring to some entity that is not yet imported (e.g. a new molecule)?  
- Are these dictionaries mandatory or optional? What will the validator do with them?
- Should they always adhere to the given format in case they are present? 
- Do these dictionaries depend on the experiment type (OP or FF)?

PR #355, opened 2025-10-07
Allow option in run_analysis to exit with non zero exit code

### LINKED ISSUE #354: Allow run_analysis to exit with non-zero code on analysis errors
Currently [run_analysis](https://github.com/NMRLipids/Databank/blob/764af3ab97012c51534fe40e3782e7d5bcd06337/src/DatabankLib/utils.py#L15) logs how many analyses failed, but the process still exits 0 even with failures. That’s fine for local or interactive use, but continuous-integration jobs benefit from being able to detect failures automatically.

MagnusSletten:
A quick proposed solution for allowing CI jobs to fail properly when analysis reports errors. Closes #354 

Maybe there are more direct ways to archive the same while touching less code, but a change like this is great for adding new simulations where we don't want to read through logs to check for errors. 

----
📚 Documentation preview 📚: https://Databank--355.org.readthedocs.build/en/355/

### DIFF
--- a/src/DatabankLib/utils.py
+++ b/src/DatabankLib/utils.py
@@ -28,6 +28,11 @@ def run_analysis(
                      (None, None) which means all systems. Can be also (None, -1)
                      which means all new systems, or (0, None) which means all
                      old systems.
+    Environment variables:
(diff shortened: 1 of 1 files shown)


MagnusSletten:
Using an environmental variable within run_analysis instead would touch less less code so that's a fine alternative.

PR #359, opened 2025-10-14
Redownload with filesize check

batukav:
This pull request adds the following functionality to `databankLibrary.py`:

1- `_download_and_verify` function that checks the size of the target files after downloading and retries the download three times if the sizes don't match,
2- progress bar for the file download

First point is necessary as the large downloads from the Zenodo can get interrupted and currently we don't have a way to catch such interruptions. 

----
📚 Documentation preview 📚: https://Databank--359.org.readthedocs.build/en/359/

### DIFF
--- a/src/DatabankLib/databankLibrary.py
+++ b/src/DatabankLib/databankLibrary.py
@@ -301,17 +301,15 @@ def system2MDanalysisUniverse(system):  # noqa: N802 (API name)
             top_name = None
         else:
             top_name = os.path.join(system_path, top)
-    except Exception as e:
-        logger.error("Error getting structure/topology/trajectory filenames.")
-        logger.error(str(e))
+    except Exception:
+        logger.exception(f"Error getting structure/topology/trajectory filenames for system {system['ID']}.")
         raise
 
     # downloading trajectory (obligatory)
     if skip_downloading:
         if not os.path.isfile(trj_name):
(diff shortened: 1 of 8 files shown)


src/DatabankLib/databankLibrary.py
```diff
@@ -275,6 +277,83 @@ def calc_z_dim(gro):
...
+            Size of each block (in tqdm units) [default: 1].
+        tsize  : int, optional
+            Total size (in tqdm units). If [default: None] remains unchanged.
+        """
+        if tsize is not None:
+            self.total = tsize
+        self.update(b * bsize - self.n)  # will also set self.n = b * bsize
+
+
+def _download_and_verify(url: str, dest_path: str, expected_size: int | None = None):
```
comcon1: You can move the whole function to `databankio` because download_and_verify seems to be quite a general useful function. Make it also non-private there.

src/DatabankLib/databankLibrary.py
```diff
@@ -275,6 +277,83 @@ def calc_z_dim(gro):
     return z
 
 
+class TqdmUpTo(tqdm):
```
comcon1: It doesn't look like this class should be public.

comcon1 (changes requested):
suggestion to compactize

PR #368, opened 2025-11-03
Metadata enhance

korbinib:
parallel to https://github.com/NMRLipids/BilayerData/pull/110

----
📚 Documentation preview 📚: https://fairmd-lipids--368.org.readthedocs.build/en/368/


developer/AutocompleteMetadata.py
```diff
@@ -0,0 +1,221 @@
+import os
+import sys
+import requests
+import yaml
```
comcon1: Please, add the comment clarifying what scripts does on the top of the file.

comcon1 (changes requested):
@korbinib could you please also move files according to last folder reorganisation?

developer/AutocompleteMetadata.py
comcon1: Yes!

Except that comment to the entire file should stay in the very top before imports. 

Also you can directly refer to the page in the documentation (https://nmrlipids.github.io/FAIRMD_lipids/stable/dbstructure.html#molecule-metadata). This subchapter is not filled yet (but should one day).

```
.. _molecule_record:

Molecule record
---------------

Molecule metadata
~~~~~~~~~~~~~~~~~
TODO: write the block

```

So, you can refer to "Molecule record". By using 
```
:ref:`molecule_record`
```
or 
```
:ref:`molecule representation in the databank <molecule_record>`
```
if you want to specify text for the link. 

We do parse `developer` folder (https://nmrlipids.github.io/FAIRMD_lipids/stable/auto_gen/developer.html) and it's not good that all pages for the files there are empty.

PR #377, opened 2025-11-07
Improve add data behavior together with corresponding important API additions

comcon1:
Here is important modifications for add_simulation behavior. Now, it explicitly points, were is the problem in the mapping file if there is a problem. Previously we just compared atom numbers which does'nt actually help a lot when you made a type in a mapping file. 

1. @MagnusSletten we told recently about `--non-interactive` behavior of add_data. Here it is.
2. @hsantila Here are important additions. Methods `nml2selection` and `md2nml` are introduced together with tests. Also, important function `check_mapping` allow now to test mapping dictionary against MDAnalysis universe.
Note that I want to **completely remove** outdated  functions `getUniversalAtomName` and `simulation2universal_atomnames`. Also, it's a first steps to remove `getAtoms`, `getLipids`, and other similar stuff.

----
📚 Documentation preview 📚: https://fairmd-lipids--377.org.readthedocs.build/en/377/

### DIFF
--- a/src/fairmd/lipids/settings/molecules.py
+++ b/src/fairmd/lipids/settings/molecules.py
@@ -37,14 +37,14 @@ class MoleculeError(Exception):
 class MoleculeMappingError(MoleculeError):
     """Custom exception for molecule mapping errors."""
 
-    def __init__(self, message: str, mol=None, *args, **kwargs) -> None:
+    def __init__(self, message: str, mol=None) -> None:
         if mol is None:
             msg = message
         elif mol.mapping_file is None:
             msg = f"From {mol}: {message}"
         else:
             msg = f"From {mol}[{mol.mapping_file}]: {message}" if mol is not None else message
-        super().__init__(msg)
+        super().__init__(msg, mol=mol)
 
 
 class Molecule(ABC):
@@ -85,6 +85,8 @@ class Molecule(ABC):
 
         :param u: MDAnalysis Universe
         :param name: string standing for residue name if it is not in the mapping file
+
+        :raises MoleculeMappingError: if mapping is inconsistent
         """
         res_set = {v["RESIDUE"] for _, v in self.mapping_dict.items() if v.get("RESIDUE") is not None}
         res_set.add(name)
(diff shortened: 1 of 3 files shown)


src/fairmd/lipids/settings/molecules.py
```diff
@@ -46,16 +48,63 @@ def register_mapping(self, fname: str) -> None:
...
+        for a in res_atoms:
+            _ = self.md2uan(a.name)
+        for unm in self.mapping_dict:
+            sel_str = self.uan2selection(unm, name)
+            ats: mda.AtomGroup = u.select_atoms(sel_str)
+            if ats.n_atoms == 0:
+                msg = f"Atom {unm} was not found in the universe using selection '{sel_str}'."
+                if name in sel_str:
+                    msg += f" Could be mapping error or incorrect residue name '{name}'."
+                raise KeyError(msg)
```
MagnusSletten: Maybe it's worth to have a custom MappingError exception here because KeyError can appear in many places in add_simulation.py or the converted "RuntimeError ". If I catch MappingError I would be sure about what I am actually catching. 

This practice isn't really done anywhere else in the code though.

MagnusSletten (commented):
This is great! Very happy about these changes.  I’m unfortunately a bit busier than I expected right now, so it’ll take me a little longer to go through the code.

src/fairmd/lipids/settings/molecules.py
comcon1: Is it enough for now you think?
See commit f9fb636

MagnusSletten (commented):
The new Mapping related exceptions looks great. Will be much easier to handle now 👍

src/fairmd/lipids/settings/molecules.py
```diff
@@ -7,17 +7,46 @@
...
+    """Custom exception for molecule mapping errors."""
+
+    def __init__(self, message: str, mol=None, *args, **kwargs) -> None:
+        if mol is None:
+            msg = message
+        elif mol.mapping_file is None:
+            msg = f"From {mol}: {message}"
+        else:
+            msg = f"From {mol}[{mol.mapping_file}]: {message}" if mol is not None else message
+        super().__init__(msg)
```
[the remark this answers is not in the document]
comcon1: I've fixed it

Issue #395, opened 2025-12-08
Form-factor of different samples - to distinguish


comcon1:
10.1103.008
there two form factors WITH ABSOLUTELY IDENTICAL README.yaml
https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.100.198103

but the SAMPLES ARE VERY DIFFERENT.

This question is likely related to this PR: https://github.com/NMRLipids/Databank/pull/352

comcon1:
Will be addressed in #393 soon.

PR #396, opened 2025-12-08
Modify experiment metadata to track path and not DOI

### LINKED ISSUE #394: Use path in EXPERIMENT field instead of DOI: PATH
Some experiment records are already without DOI field in YAML. We should use path instead.

comcon1:
We start using path (`10.0021/cp34253245/1`) instead of just DOI which is ambiguous. It closes #394 
- match_experiments is modified to track only DOI
- metadata changed: now we have lists instead of dicts
```yaml
EXPERIMENT:
   ORDERPARAMETER:
     - pathA
     - pathB
   FORMFACTOR:
     - path1
     - path2
```
- multiple formfactor experiments are now allowed
- we take the best formfactor quality as one we stored. We are not averaging currently, which is bad. So we adressed #395 only partly

----
📚 Documentation preview 📚: https://databank--396.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/bin/evaluate_quality.py
+++ b/src/fairmd/lipids/bin/evaluate_quality.py
@@ -56,10 +56,8 @@ def evaluate_quality():
 
             for path in simulation.system["EXPERIMENT"]["ORDERPARAMETER"][lipid1]:
                 print(
-                    f"Evaluating {lipid1} lipid using experimental data from{FMDL_EXP_PATH}/OrderParameters/{path}",
+                    f"Evaluating {lipid1} lipid using experimental data from {FMDL_EXP_PATH}/OrderParameters/{path}",
                 )
-
-                print(path)
                 OP_qual_data = {}
                 # get readme file of the experiment
                 exp_fpath = os.path.join(FMDL_EXP_PATH, "OrderParameters", path)
--- a/src/fairmd/lipids/quality.py
+++ b/src/fairmd/lipids/quality.py
@@ -8,6 +8,7 @@ import decimal as dc
 import json
 import os
 import re
+from typing import Optional, Tuple
 
 import numpy as np
 import scipy.signal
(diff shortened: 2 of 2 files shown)


src/fairmd/lipids/bin/match_experiments.py
```diff
@@ -255,9 +255,12 @@ def find_pairs(experiments: list[Experiment], simulations: list[SearchSystem]):
                         exp_total_lipid_concentration / sim_total_lipid_concentration > 1 - LIP_CONC_REL_THRESHOLD
                     ) and (exp_total_lipid_concentration / sim_total_lipid_concentration < 1 + LIP_CONC_REL_THRESHOLD):
                         switch = 1
-                elif (type(exp_total_lipid_concentration) is str) and (type(sim_total_lipid_concentration) is str):
-                    if exp_total_lipid_concentration == sim_total_lipid_concentration:
-                        switch = 1
+                elif (
+                    (type(exp_total_lipid_concentration) is str)
```
batukav: what is the reason for this type checking? can we think of a better way to do this implementation? Do we have non-numerical values for the concentration in the experiment yamls?

batukav (changes requested):
I'd like to resolve my comments before accepting this PR

src/fairmd/lipids/bin/match_experiments.py
comcon1: I didn't modify something here. I just pleased ruff a bit to reduce code complexity. Unfortunately, it's a legacy. Once you are done with the experiment class, and we start to migrate to the extended metadata, it's a first thing I would like to get rid of. Currently, if n_water/n_lipid > 25, then simulation hydration is set to string "full hydration". It is also written "full hydration" in experiment YAML despite it's always possible to write a normal numeric value based on the paper. This ratio 25 is actually well only for simple phospholipids because if lipid is cardiolipin or LPS, this number "25" is different. If it's GB3 or GM1, I'm also sure it's different. I already introduced new hydration metadata and we need to decide how to process it.

batukav (approved):

PR #404, opened 2025-12-16
Prettify databanklib

comcon1:
Here I will remove unused functions, rename useful ones, clean their documentation, etc

----
📚 Documentation preview 📚: https://databank--404.org.readthedocs.build/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ dependencies = [
     "tqdm>=4.5.0",
     "periodictable>=1.5.0",
     "Deprecated",
+    "requests",
     "scipy",
     "jsonschema",
 ]


src/fairmd/lipids/SchemaValidation/validate_info_dict.py
hsantila: The rows 40-43 are a bit hard to decipher. Maybe something like this is what you're after here:

missing_keys = [
    k for k in software_required_keys
    if k not in sim or sim[k] is None
]

if missing_keys:
    software = sim.get("SOFTWARE", "<unknown>")
    msg = (
        f"Required '{software.upper()}' sim keys missing or "
        f"not defined in conf file: {', '.join(missing_keys)}"
    )
    raise YamlBadConfigError(msg)

src/fairmd/lipids/api.py
hsantila: I think the option for "mass" is unimplemented for now for membrane_composition, unlike what the documentation implies?

hsantila (changes requested):
Here are some preliminary comments, probably didn't catch all things but hoping the keen eyes of other reviewers will:)

comcon1:
I've rebased the branch before satisfying review.

src/fairmd/lipids/api.py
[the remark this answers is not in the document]
comcon1: I don't see any use of a Show-function if you have Get-function. Show-function everyone can easily write on his own.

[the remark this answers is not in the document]
comcon1: Of course, we don't. `quality.py` is a half-spagetti, this PR doesn't include `quality.py` revision. But that we can do, it make sence. Should we replace the function in `api.py` with the quality-py's function?

src/fairmd/lipids/SchemaValidation/validate_info_dict.py
```diff
@@ -0,0 +1,139 @@
...
+                            [f.strip()] for f in value_sim_splitted
+                        ]  # IMPORTANT: Needs to be list of lists for now
+                    files_tbd.extend(f[0] for f in sim[key_sim])
+
+                # STEP 4.3.
+                # Batuhan: In conf file only one psf/tpr/pdb file allowed each
+                # (can coexist), multiple TRJ files are ok
+                # TODO true for all sim software?
+                # TODO add dict entry param "unique" instead?
+                if key_sim.upper() in ["PSF", "TPR", "PDB"] and len(sim[key_sim]) > 1:
```
batukav: I am not sure if this implementation is in line with the comments:

`
                # STEP 4.3.
                # Batuhan: In conf file only one psf/tpr/pdb file allowed each
                # (can coexist), multiple TRJ files are ok
                # TODO true for all sim software?
                # TODO add dict entry param "unique" instead?
`

Maybe we should enforce a valid topology file and remove PDB from the accepted ones?

src/fairmd/lipids/core.py
```diff
@@ -81,14 +82,74 @@ def content(self) -> dict[str, Molecule]:
...
+        for k, v in self["COMPOSITION"].items():
+            if k not in lipids_set:
+                continue
+            count = sum(v["COUNT"])
+            comp[k] = count
+        n_lipids = self.n_lipids
+        if which == "molar":
+            for k in comp:
+                comp[k] /= n_lipids
+        elif which == "mass":
```
batukav: if we are planning to support further measures other than `molar` and  `mass`, fine to keep it like this. otherwise, `elif` -> `else` as there are only two options anyways.

src/fairmd/lipids/core.py
```diff
@@ -81,14 +82,74 @@ def content(self) -> dict[str, Molecule]:
...
+                mw = mol.metadata["bioschema_properties"]["molecularWeight"]
+                comp[k] *= mw
+                total_mass += comp[k]
+            for k in comp:
+                comp[k] /= total_mass
+        return comp
+
+    def get_hydration(self, which: str = "number") -> float:
+        """Get system hydration."""
+        if which == "number":
```
batukav: let's make this if/else consistent with the `membrane_composition` method. -> 
```
        if which not in ["molar", "mass"]:
            msg = "Which must be 'molar' or 'mass'"
            raise ValueError(msg)
```

that is, check the validity of hydration metric before the if/else. (or change the way it is done in `membrane_composition`)

src/fairmd/lipids/SchemaValidation/validate_info_dict.py
```diff
@@ -0,0 +1,139 @@
...
+
+from fairmd.lipids.molecules import lipids_set, molecule_ff_set, molecules_set
+from fairmd.lipids.SchemaValidation.engines import software_dict
+
+
+class YamlBadConfigError(Exception):
+    """Custom Exception class for parsing the yaml configuration"""
+
+
+def parse_valid_config_settings(info_yaml: dict, logger) -> tuple[dict, list[str]]:
```
batukav: @MagnusSletten is this the same validator you use for the upload portal? 

@comcon1 apologize if I am missing something, but do we validate the entire config somewhere else, or this is where we are supposed to do it? 

if this is the only place, then I strongly suggest we validate each entry, including their type. For instance, molecule counts have to be integers etc.

comcon1: We have also a schema and it should verify types. That @korbinib made. But this question is a bit out of this PR. It appears in diffs just because I moved it from `api.py` because schema-related stuff are somehow gathered in this folder `SchemaValidation`. Which is by some reason in CamelCase that we should avoid. So there are some questions about validation and we can make a PR about it on the next step.

src/fairmd/lipids/SchemaValidation/validate_info_dict.py
batukav: This could be a mistake, indeed. 

Depending on how many systems are affected by this, we can either contact the authors to upload topologies, or try to rerun the systems on our own.

src/fairmd/lipids/SchemaValidation/validate_info_dict.py
MagnusSletten: I use both this file and the [json-schema I added](https://github.com/NMRLipids/FAIRMD_lipids/blob/f186a8d5d8946db01e0a176f335b71d5efc4255b/src/fairmd/lipids/SchemaValidation/Schema/info_yml_schema.json). There's some overlap in what they check and the json-schema is a more standard rigorous way to check the files @batukav 

It has some limitations though so maybe more domain-information is checked with other files like this one. I'm doing some changes related to a new README.yaml schema I'm adding and I'll including changing the filenames then. The camel-case has been annoying me every time I see it 😄

src/fairmd/lipids/SchemaValidation/validate_info_dict.py
```diff
@@ -0,0 +1,139 @@
...
+                if key_sim.upper() in ["PSF", "TPR", "PDB"] and len(sim[key_sim]) > 1:
+                    msg = f"only one '{key_sim}' entry file allowed, but got {len(sim[key_sim])}: {sim[key_sim]}"
+                    raise YamlBadConfigError(msg)
+
+        else:
+            logger.warning(
+                f"skipping key '{key_sim}': Not defined in software_sim library",
+            )
+
+    logger.info(f"found {len(files_tbd)} resources to download: {', '.join(files_tbd)}")
```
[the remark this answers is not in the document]
comcon1: Done

src/fairmd/lipids/SchemaValidation/validate_info_dict.py
comcon1: Issue #413 is created. Resoliving here.

src/fairmd/lipids/core.py
```diff
@@ -81,14 +82,74 @@ def content(self) -> dict[str, Molecule]:
...
+    @property
+    def n_lipids(self) -> int:
+        """Returns total number of lipid molecules in the system."""
+        total = 0
+        for k, v in self["COMPOSITION"].items():
+            if k in lipids_set:
+                total += sum(v["COUNT"])
+        return total
+
+    def membrane_composition(self, which: str = "molar") -> dict[str, float]:
```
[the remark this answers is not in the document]
comcon1: I renamed to "basis" and wrote this hint:
```
basis: typing.Literal["molar", "mass"] = "molar"
```

src/fairmd/lipids/api.py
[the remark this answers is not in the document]
hsantila: yup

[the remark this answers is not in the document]
comcon1: @hsantila I started satisfying this one. It requires some starting architectural changes. I isolate FF-curve-processing functions into a separated file. Please follow the last commits.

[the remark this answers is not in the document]
comcon1: See 17f7bc95 and above.
I removed this function **COMPLETELY** from api. Instead, I add get_FF to api.py and get_mins_from_ffdata to `analib/formfactor.py`. Now quality-file uses this function.

PR #411, opened 2026-01-05
Drafting PR for revision of the ranking

comcon1:
- [x] definition of get_quality API
- [x] implementing get_quality API
- [x] document get_quality and quality formulations
- [x] rewritting make_ranking to form normal-looking tables using get_quality API

Will fix issue #412.

Here, one can see how the new ranking will look like in GitHub: https://github.com/NMRLipids/BilayerData/blob/view-new-rankings/Ranking/FF_ranking.csv
I think, it's quite nice.

----
📚 Documentation preview 📚: https://databank--411.org.readthedocs.build/

### DIFF
--- a/docs/src/dbstructure.rst
+++ b/docs/src/dbstructure.rst
@@ -183,6 +183,26 @@ more information, see :ref:`molecule_names`
 Other files
 -----------
 
+.. _ranking_files:
+
+Ranking files
+~~~~~~~~~~~~~~~~~
+Contain tables where major conditions and quality scores are stored for each simulation in CSV format. These files are stored at ``Ranking`` subfolder of :py:data:`fairmd.lipids.FMDL_DATA_PATH` folder. We store:
+
+1. Form-factor ranking files named as ``FF_ranking.csv``.
+2. Order paparmeter ranking files named as ``OP_ranking.csv``.
+3. Individual lipid ranking files named as ``<lipid>_ranking.csv``.
+
+The difference between 2. and 3. is that 2. contains average quality scores over all lipids in the simulation, while 3. contains quality scores for only selected lipid.
+
+.. _uadic_files:
+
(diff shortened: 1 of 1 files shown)


ohsOllila:
Idea looks reasonable to me.

>     * there is currently no quality combined between experiments (but we are free to define and we should define here).

What do you mean with this?
 
>     * there is "system" quality which is a combination of all the lipids' particular qualities (which is IMHO no use to store because we always can easily compute from fragment-files). I would like to remove all SYSTEM quality files.

If the function returns this and total rankings are somehow accessible, then I agree that they can be removed.

>     * there are fragment qualities per lipid type, which is a combination of atomic qualities. That is good to have despite we should remember that a lot of experiments with multiple lipids has experimental values covering many different lipid types if they have similar fatty acids. But that we cannot cover here except mentioning in the documentation.

Yes, this depends how experimental data is assigned when put in the databank, and this is the decision of person who is adding it.

> There is no documentation describing how quality is computed and merged which is **VERY BAD** because quality is kind of a feature.

Quality calculation is explained in the supplementary information of the publication.

ohsOllila:
> _Regarding the first question:_ We do not combine FF and OP into one value, but we should. At least, I don't see any traces of it in the code. But we should. We also have problems combining more than one FF experiment. We have one case already.

Yes, these are not combined. It requires ad hoc decision on normalizing and weighting them. I decided not to do this previously. Partially because I believe that tail order parameters are measuring membrane packing quite well.

comcon1:
Ok. I will try to do it if I will have energy. I have some thoughts.

Best regards, Alex

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

batukav (approved):
Good work. Looks good to me.

PR #414, opened 2026-01-07
fix license key and allow py314

comcon1:
small fix in license info

----
📚 Documentation preview 📚: https://databank--414.org.readthedocs.build/


comcon1:
I've edited history to make 3 clear commits.

PR #425, opened 2026-01-15
Resumable download

comcon1:
- make file.part and AFTER downloading, renames
- can tries with resuming download OVER urllib3 retries (high-level retry)

----
📚 Documentation preview 📚: https://databank--425.org.readthedocs.build/


comcon1:
I am actually use this branch for real recomputing and try to collect real exceptions which can be issued. I am suspecting that ReadTimeout is not the only one.

src/fairmd/lipids/databankio.py
```diff
@@ -110,40 +123,58 @@ def download_with_progress_with_retry(
...
+        desc=tqdm_title,
+    ) as u:
+        # go
+        if os.path.isfile(dest) and dest.endswith(".part"):
+            # Resuming download
+            dl_size = os.path.getsize(dest)
+
+            n_chunks = dl_size // chunk_sz
+            downloaded = chunk_sz * n_chunks
+            if dl_size % chunk_sz != 0:
```
MagnusSletten: Why do we do truncation here?

src/fairmd/lipids/databankio.py
```diff
@@ -110,40 +123,58 @@ def download_with_progress_with_retry(
...
+                    f.truncate(downloaded)
+            u.update_retrieve(b=downloaded, bsize=1, tsize=total_size)
+            mode = "ab"
+            headers = {"Range": f"bytes={downloaded}-"}  # request only missing part
+        else:
+            mode = "wb"
+            headers = {}
+            downloaded = 0
+        # open connection
+        with open(dest, mode) as f, _open_url_with_retry(uri, update_headers=headers) as resp:
```
MagnusSletten: Should we verify resp.status_code is 206 when resuming? If the server returns 200, the range was ignored and appending can corrupt the file.
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#206

MagnusSletten (commented):
Very nice pr! Just a couple of comments. We can also test this quite fast immediately after merge with a sim to check everything. I do still get some errors on the full test suite that looks similar to last full run via actions here.

src/fairmd/lipids/databankio.py
comcon1: These error appears not in this PR. All except one come from enabling py314. One comes from changing schema for NAMD processing. Let's make an issue out of it but will fix not here.

comcon1:
I rebased above the large PR with Experiment class

src/fairmd/lipids/databankio.py
```diff
@@ -118,32 +134,63 @@ def update_retrieve(self, b=1, bsize=1, tsize=None):
...
+        desc=tqdm_title,
+    ) as u:
+        # go
+        if os.path.isfile(dest):
+            # Resuming download
+            dl_size = os.path.getsize(dest)
+
+            n_chunks = dl_size // _fmdl_chunk_size
+            downloaded = _fmdl_chunk_size * n_chunks
+            if downloaded != dl_size:
```
batukav: If this is true, we exit the code at line 156, so we are not actually continuing the download. Sounds contradictory to what we are actually trying to achieve, am I right?

comcon1:
image
It's a real work of the algorithm

PR #427, opened 2026-01-16
Round quality values to 4 decimal places in output files

### LINKED ISSUE #416: QualityEvaluation stores too high precision
Machine precision affects the 8'th+ digit of quality and make DIFF in git with updated values. It's very bad because we must check by eyes whether quality ACTUALLY was changed. 

**Solution:** we should round quality in the output at 4th digit. We don't trust the 8th digit anyway.

jagdep-singh:
This PR rounds quality values to 4 decimal places at output time to avoid noisy diffs caused by floating-point precision. Rounding is done just before writing data to the file, made a function `round_quality_values(obj, ndigits=n)`.

Also saw that `system_qual_output` can have NaN values. Earlier the check was directly doing `> 0`, which is not safe in for NaNs. So changed the `SQout` logic to check for valid float values before comparing, else everything is same .... actually i commented and added the change idk why but felt it would be better 

Fixes #416

----
📚 Documentation preview 📚: https://databank--427.org.readthedocs.build/

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -20,7 +20,6 @@ Since June 2025 (v.1.1.0), FAIRMD Lipids separates codespace from [Data](https:/
 - Jagdeep Singh (@jagdep-singh)
 - Hanne Susanna Antila (@hsantila)
 - Hector Martinez-Seara Monne (@hseara)
-- Jagdeep Singh(@jagdep-singh)
 - Lara Dautzenberg (@larabort)
 - Magnus-Elias Sletten (@MagnusSletten)
 - Markus Miettinen (@markussmiettinen)


comcon1:
@jagdep-singh please add yourself to authors in this PR. We will squash commits for small PRs.

src/fairmd/lipids/bin/evaluate_quality.py
```diff
@@ -25,6 +25,29 @@
 from fairmd.lipids.auxiliary.jsonEncoders import CompactJSONEncoder
 
 
+def round_quality_values(obj, ndigits=4):
```
comcon1: - docscring must be added. 
- make function private (just add underscore _round_quality_values)
- add type hints

src/fairmd/lipids/bin/evaluate_quality.py
```diff
@@ -145,11 +171,14 @@ def evaluate_quality():
...
         outfile2 = os.path.join(DATAdir, "SYSTEM_quality.json")
-        SQout = False
-        for SQ in system_qual_output:
-            if system_qual_output[SQ] > 0:
-                SQout = True
+        # SQout = False
+        # for SQ in system_qual_output:
+        #     if system_qual_output[SQ] > 0:
+        #         SQout = True
+        SQout = any(isinstance(v, float) and not np.isnan(v) and v > 0 for v in system_qual_output.values())
```
comcon1: expression `np.nan > 0` is false. So, you don't need to test for it. Also, you don't need to test type. If there is something nonint, it should probably raise exception because it means that there is some severe error before.

comcon1 (changes requested):
That's a good way to do as you did. Thanx.  Some changes are requested.

AUTHORS.md
```diff
@@ -17,8 +17,10 @@ Since June 2025 (v.1.1.0), FAIRMD Lipids separates codespace from [Data](https:/
 - Cedric Mingham (@thraindk)
 - Fabián Suárez-Lestón (@fsuarezleston)
 - Julianne Hoeflich (@JHoeflich1)
+- Jagdeep Singh (@jagdep-singh)
 - Hanne Susanna Antila (@hsantila)
 - Hector Martinez-Seara Monne (@hseara)
+- Jagdeep Singh(@jagdep-singh)
```
comcon1: dublicate line

PR #434, opened 2026-01-18
Fix align-mappings after some API improvements

comcon1:
small fixes. I run it through the repo, seems that it works.

----
📚 Documentation preview 📚: https://databank--434.org.readthedocs.build/

### DIFF
--- a/developer/align_mapping.py
+++ b/developer/align_mapping.py
@@ -8,6 +8,8 @@ files, SMILES, and add the alignments for only heavy atoms. Further should
 be converted into a mature CI/CD script.
 """
 
+import argparse
+import os
 import re
 import sys
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


developer/align_mapping.py
```diff
@@ -105,11 +69,10 @@ def find_uname(x, atom_name, res_name) -> str:
...
 
-    with open("done.txt") as fd:
-        done_ids = list(map(int, fd.readlines()))
-
-    # list of deeply defective systems
-    done_ids += [151]
+    done_ids = []
+    # with open("done.txt") as fd:
+    #    done_ids = list(map(int, fd.readlines()))
+    done_ids += [151] # defective systems
```
batukav: `done_ids = []
done_ids.append(151)
`
feels more natural.

PR #438, opened 2026-01-19
Fixed rounding of FormFactorQuality when result is a tuple

### LINKED ISSUE #416: QualityEvaluation stores too high precision

jagdep-singh:
fixes FormFactorQuality rounding in `fmdl_evaluate_quality`.
The form factor quality returned by get_ffq_scaling is a tuple, so `_round_quality_values` was a no-op. The tuple is now converted to a list before rounding and writing `FormFactorQuality.json`

fixes #416 

----
📚 Documentation preview 📚: https://databank--438.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/bin/make_ranking.py
+++ b/src/fairmd/lipids/bin/make_ranking.py
@@ -16,7 +16,6 @@ No arguments are needed.
 
 import math
 import os
-from pathlib import Path
 
 import numpy as np
 import pandas as pd


comcon1 (changes requested):
Form-factor quality is rounded properly now. 

Also, let's finish with API function.

System-quality construction doesn't fix the problem. I was talking about this system: `68d/39e/68d39e96aca23aeaa4ea2567e0e0cb8b8903a694/68d39e96aca23aeaa4ea2567e0e0cb8b8903a694`
Here fragment quality for one tail was divided by 2.
However, this we need to design in the equation. I see now, that there are severe problems in the equation itself and we need to consult other project contributors regarding this first. Let's revert `quality.py` to its original state and complete other things in this PR.

jagdep-singh:
Thanks for the clarification. I reverted `quality.py` fully to its previous state
I simplified the API tail handling using `np.nanmean`, updated tail matching, and adjusted directory creation to match the project.

PR #439, opened 2026-01-20
crash handelling in `systemQuality` when tails data is missing

jagdep-singh:
This PR fixes a crash in `fmdl_evaluate_quality` caused by tails being `None` in systemQuality.
When tail data is unavailable, it now skips the tails instead of multiplying `None` with a float.

fixes #438 

----
📚 Documentation preview 📚: https://databank--439.org.readthedocs.build/


jagdep-singh:
will bw closing this one then

PR #448, opened 2026-01-23
Nmrpca fix egligibility

comcon1:
Now, all the lipids having sn-1 and sn-2 tails are egligible. List is not hardcoded anymore.

----
📚 Documentation preview 📚: https://databank--448.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/analib/maicos.py
+++ b/src/fairmd/lipids/analib/maicos.py
@@ -25,7 +25,7 @@ from maicos.lib.util import get_compound
 from maicos.lib.weights import density_weights
 from tqdm import tqdm
 
-from fairmd.lipids.auxiliary.jsonEncoders import CompactJSONEncoder
+from fairmd.lipids.auxiliary import CompactJSONEncoder
 from fairmd.lipids.core import System
 from fairmd.lipids.molecules import lipids_set
 
--- a/src/fairmd/lipids/analyze.py
+++ b/src/fairmd/lipids/analyze.py
@@ -43,8 +43,7 @@ from fairmd.lipids.analib.maicos import (
     traj_centering_for_maicos_mda_parallel,
 )
 from fairmd.lipids.api import UniverseConstructor, mda_gen_selection_mols
-from fairmd.lipids.auxiliary import mollib
-from fairmd.lipids.auxiliary.jsonEncoders import CompactJSONEncoder
+from fairmd.lipids.auxiliary import CompactJSONEncoder, mollib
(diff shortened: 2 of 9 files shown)


src/fairmd/lipids/analib/analyze_nmrpca.py
```diff
@@ -288,19 +216,33 @@ def prepare_OpenMM_traj(self):  # noqa: N802
...
-        """
-        Create Concatenator and corresponding concatenated trajectories for
-        all lipids available for the trajectory.
+            self._prepare_gmx_traj()
+
+    @staticmethod
+    def verify_lipid(lipid: Lipid) -> bool:
+        """Verify that the lipid is supported by the method."""
+        __tailsn1_vars = ["sn-1", "sn1"]
+        __tailsn2_vars = ["sn-2", "sn2"]
```
pbuslaev: Should this be stored elsewhere and only imported here? I believe that tail-names are universal and not class-specific.

pbuslaev: Also, I believe that this method can hardly be generalized when 4-tailed lipids become available. Should the `Parser` class have a dict of supported tail combinations

```python
from collections import Counter
from ... import TAIL_VARS, REVERSE_TAIL_VARS  # dictionary of tail alt-names
SUPPORTED_LIPIDS = {
    2: ["sn1", "sn2"],  # List of tail names for lipids with 2 tails
}

@classmethod
def verify_lipid(cls, lipid: Lipid) -> bool:
    tails = []
    for v in lipid.mapping_dict.values():
        if v["FRAGMENT"] in REVERSE_TAIL_VARS:
            tails.append(REVERSE_TAIL_VARS[v["FRAGMENT"]])
    if len(tails) not in SUPPORTED_LIPIDS:
        return False
    return Counter(tails) == Counter(SUPPORTED_LIPIDS[len(tails)])
"""

File: src/fairmd/lipids/analib/analyze_nmrpca.py [hunk 2]
```diff
@@ -336,15 +238,17 @@ class Topology:
     :param mapping_dict: preloaded mapping_dict
     """
 
+    HEADGRP = "headgroup"  # locally used to gather amber lipids together
+    TAILSN1 = "sn-1"
+    TAILSN2 = "sn-2"
```
pbuslaev: @comcon1 As you've mentioned, this is very specific for Amber and two tails. Should we add a `NOTE` or `TODO` to mark this bit to improve in the future?

Issue #420, opened 2026-01-13
Generate JSON for OP plotting


comcon1:
Generate exportable JSON for OP plotting covering all lipid types.

PR #451, opened 2026-01-27
Refactoring base classes, OP dictionary, match_experiments

comcon1:
## Refactoring
`_base.py` is created with some base classes. SampleComposition interface is separated and both Experiment and System implements it.

## match_experiments 
It is completely refactored and comparators are separated into a list of matcher-functions. These functions can be easily tested separately which is **VERY** important.

## Nicely looking OP dictionary
We decided together with @mdondrup that we will form the nice OP dict of the form:
```json
{
    "fragment-1": [
        { "C": "C1", "H": "H11", "OP": -0.18, "err": 0.01},
        { "C": "C1", "H": "H12", "OP": -0.17, "err": 0.02},
        { "C": "C2", "H": "H2X", "OP": -0.05, "err": 0.02},
    ],
    "fragment-2": {
    }
}
```
It should work for both experiment and simulation. I am doing some small code prettification and small refactoring and minimalizing core code while doing this job because it will interact a lot with Experiment class and its API is not very well established still.

This will solve PR #420.

----
📚 Documentation preview 📚: https://databank--451.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/molecules.py
+++ b/src/fairmd/lipids/molecules.py
@@ -13,6 +13,7 @@ import re
 from abc import ABC, abstractmethod
 from collections.abc import MutableSet
 from copy import copy
+from glob import glob
 from typing import Any
 
 import MDAnalysis as mda
@@ -77,11 +78,8 @@ class Molecule(ABC):
         """
         # iterate over possible paths
         if fname is None:
-            _possible_mfiles = [
-                f
-                for f in os.listdir(self._get_path())
-                if f.endswith((".yaml", ".yml")) and not f.startswith("metadata")
-            ]
+            path = self._get_path()
+            _possible_mfiles = [os.path.basename(f) for f in glob(os.path.join(path, "mapping*.y*ml"))]
             if len(_possible_mfiles) == 0:
                 msg = f"No mapping file found in {self._get_path()}"
                 raise MoleculeMappingError(msg, mol=self)


mdondrup (changes requested):
I think the format is ok for plotting. I have tried to get the data though. It isn't clear how to access this. A docstring needs to be added to the ExperimentOP class that demonstrates how to access the OP data per lipid in the experiment. Please implement a getter method in this class.

comcon1:
Thanks for review. Will work on docstrings. What do you mean by getter? There is 'data' property in the class which returns data dictionary.

Best regards, Alex

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

src/fairmd/lipids/bin/match_experiments.py
```diff
@@ -21,210 +21,105 @@
...
+# Helper functions for matching criteria
+# TODO: remove to a module and TEST!
+
+
+def _match_membrane_composition(sim_mf: dict, exp_mf: dict) -> bool:
+    """Compare two membrane compositions given as molar fractions."""
+    if set(sim_mf.keys()) != set(exp_mf.keys()):
+        return False
+    abs_tolerance_molfraction = 0.03
+    # TODO: BAD! use relative threshold instead!
```
batukav: can we fix this todo during this PR?

batukav (changes requested):
It looks pretty good and consistent to me. I have couple of minor comments, then we can merge.

src/fairmd/lipids/molecules.py
```diff
@@ -67,13 +68,25 @@ def _get_path(self) -> str:
...
+        if fname is None:
+            _possible_mfiles = [
+                f
+                for f in os.listdir(self._get_path())
+                if f.endswith((".yaml", ".yml")) and not f.startswith("metadata")
+            ]
+            if len(_possible_mfiles) == 0:
+                msg = f"No mapping file found for in {self._get_path()}"
+                raise MoleculeMappingError(msg, mol=self)
+            fname = _possible_mfiles[0]  # take the first one
```
mdondrup: Isn't there a convention for mapping file names that they should start with `mapping`? Also, molecules like POPC have many mapping files.

tests/test_experiment.py
```diff
@@ -205,13 +236,25 @@ def mock_exp_data_path(tmpdir, monkeypatch):
 
     # Create dummy OP experiment
     with open(op_path.join("README.yaml"), "w") as f:
-        yaml.dump({"TEMPERATURE": 310.0}, f)
+        yaml.dump(
+            {
+                "TEMPERATURE": 310.0,
+                "MOLAR_FRACTIONS": {"POPE": 1.0},
```
mdondrup:
```suggestion
                "MEMBRANE_COMPOSITION": {"POPE": 1.0},
```

Issue #431, opened 2026-01-18
Non-interactive TQDM


comcon1:
TQDM should be improved to acquire TQDM_NONINTERACTIVE. Only after PR #425 will be merged.

This is a code snippet demonstrating how a non-interactive version of TQDM can be settled and used.

```python=
class LoggingTqdm(tqdm):
    def __init__(self, *args, logger=None, level=logging.INFO, **kwargs):
        self.logger = logger or logging.getLogger(__name__)
        self.level = level
        super().__init__(*args, **kwargs)

    def display(self, msg=None, pos=None):
        """Redirect tqdm's display() to the logger."""
        proc = (self.n / self.total)*100
        self.logger.log(self.level, "%8.2f%% downloaded" % proc)

pbar = LoggingTqdm(
            total=total_sz,
            desc=filename,
            unit='B',
            unit_scale=True,
            unit_divisor=1000,
            miniters=1,
            mininterval=10,
            logger=logger
        )

with requests.get(url, stream=True, headers=headers) as r, open(partial_out_path, mode) as f:
        if r.status_code not in (200, 206):
            r.raise_for_status()
        for chunk in r.iter_content(chunk_size=chunk_sz):
            f.write(chunk)
            pbar.update(len(chunk))

```

We don't have an agreed logging strategy yet, so `print(.., file=sys.stderr)` can be used instead of `logging` for now.

It would be useful to make it from the beginning in the following way. For example, in `utils.py` we could define 

```python
def fmdl_tqdm(*args, **kwargs):
    if "TQDM_NONINTERACTIVE" in os.environ:
        # do some stuff here
    else:
        return tqdm(*args, **kwargs)
```

PR #452, opened 2026-01-28
Disable tqdm in non-interactive environments

jagdep-singh:
Automatically disables tqdm when stdout is not a TTY.
Keeps progress bars in interactive terminals and avoids CI/log noise.

issue #431

----
📚 Documentation preview 📚: https://databank--452.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/molecules.py
+++ b/src/fairmd/lipids/molecules.py
@@ -13,6 +13,7 @@ import re
 from abc import ABC, abstractmethod
 from collections.abc import MutableSet
 from copy import copy
+from glob import glob
 from typing import Any
 
 import MDAnalysis as mda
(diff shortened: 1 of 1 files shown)


src/fairmd/lipids/analyze.py
```diff
@@ -172,7 +172,7 @@ def computeAPL(  # noqa: N802 (API)
         # in the databank
         apl = {}
         nlipids = system.n_lipids
-        for _ts in tqdm(u.trajectory, desc="Scanning the trajectory"):
+        for _ts in tqdm(u.trajectory, desc="Scanning the trajectory", disable=not sys.stdout.isatty()):
```
comcon1: APL is also fast, so it's also probably fine.

comcon1 (changes requested):
In places where the cycle is fast, it can be disabled. In other - not. We must have progress in logs. Progress should be redirected to stdout (the minimum). Max - it should be logged in a more log-friendly way.

PR #454, opened 2026-02-04
Parallel implementation of MDA-based centering

batukav:
----
📚 Documentation preview 📚: https://databank--454.org.readthedocs.build/

### DIFF
--- a/developer/align_mapping.py
+++ b/developer/align_mapping.py
@@ -22,13 +22,14 @@ from rdkit.Chem import MolStandardize
 
 import fairmd.lipids.api as dlapi
 import fairmd.lipids.core as dlc
+import fairmd.lipids.molecules as dlm
 from fairmd.lipids.auxiliary import elements
 
 lg = RDLogger.logger()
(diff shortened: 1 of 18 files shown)


batukav:
I'll rerun for the Drude system. let you know of the results

tests/test_maicos_parallel.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+@pytest.fixture
+def simple_universe():
+    """Create a simple MDAnalysis Universe for testing."""
+    import MDAnalysis as mda
+    from MDAnalysis.lib.mdamath import make_whole
+
+    # Use MDAnalysis test data
+    from MDAnalysis.tests.datafiles import GRO, XTC
```
comcon1: This is good to use mdanalysis test data files. But it requires mdanalysistests package to be installed (https://pypi.org/project/MDAnalysisTests/) in tox testing environment.

Please run tests which you modified. If you would run them, you will see that they fail.

PR #456, opened 2026-02-07
Increased robustness of opconvertor.py when running on some experimental data 

mdondrup:
Made build_nice_OPdict more robust in corner cases.

- OP data for some experiments is lacking standard deviation values, leading to an IndexError, added length check
- added a check wether the atom is actually contained in the mapping dict to avoid a KeyError, missing atoms are skipped
- 'nice_OPdict' contain a lot of NaN, added a function to replace those because it almost certainly will be dumped to JSON

----
📚 Documentation preview 📚: https://databank--456.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/auxiliary/opconvertor.py
+++ b/src/fairmd/lipids/auxiliary/opconvertor.py
@@ -71,7 +71,6 @@ class NamingRegistry:
                 raise ValueError(f"Unexpected H format: {row['H']}")
             idx = int(match[1])
             row["H"] = str(idx)
-           
 
             return row
 
@@ -88,7 +87,7 @@ def build_nice_OPdict(src: dict, lipid: Lipid) -> dict:
     :return: nicely formatted OP dictionary
     """
 
-    # Helper function to convert NaN to None for better 
+    # Helper function to convert NaN to None for better
     # JSON compatibility in output
     def _rnan(x: float) -> float | None:
         return None if x != x else x


tests/test_op.py
```diff
@@ -73,3 +73,78 @@ def has_c(cname: str, flist: dict) -> bool:
...
+        4. Validates the output structure and content
+        """
+        import fairmd.lipids
+        from fairmd.lipids.experiment import ExperimentCollection, ExperimentError
+        from fairmd.lipids.auxiliary.opconvertor import build_nice_OPdict
+
+        # Set the data path to the test data directory
+
+        fairmd.lipids.FMDL_DATA_PATH = "../../BilayerData"
+        experiments = ExperimentCollection.load_from_data("OPExperiment")
```
comcon1: Remove

.github/workflows/tests-min.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
     - name: run Python tests
       run: |
-        tox -e tests-min
+        tox -e tests-min -r
```
comcon1: revert

src/fairmd/lipids/auxiliary/opconvertor.py
```diff
@@ -80,15 +81,26 @@ def build_nice_OPdict(src: dict, lipid: Lipid) -> dict:
...
+    # Helper function to convert NaN to None for better JSON compatibility in output
+    def _rnan(x: float) -> float | None:
+        return None if x != x else x
+
+    def _fragmentize(src: dict, mdict: dict) -> dict:
         r = {}
         for apair, opvals in src.items():
             atom_c, atom_h = apair.split(" ")
+            if atom_c not in mdict:
+                continue
```
comcon1: raising is fine

PR #467, opened 2026-02-20
Modify apl storage

comcon1:
- changes schema of apl.json
From:
```json
{
 "0.0": 62.4,
 "20.0": 63.1
}
```
to:
```json
[
   [0, 62.4],
   [20.0, 63.1]
]
```
- remove code duplication in data-retrieving API
- add `get_density` functionality

**NOTE!** Can be merged only together with corresponding schema changes in BilayerData.

----
📚 Documentation preview 📚: https://databank--467.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/api.py
+++ b/src/fairmd/lipids/api.py
@@ -41,6 +41,20 @@ from fairmd.lipids.schema_validation.engines import get_struc_top_traj_fnames
 
 logger = logging.getLogger(__name__)
 
+__all__ = [
+    "UniverseConstructError",
+    "UniverseConstructor",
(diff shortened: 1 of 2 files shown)


batukav:
The schema would be the same for all time/spatial series data, including density profile, form factors etc.

After some thoughts, I agree that we should go with the initial schema suggestion with the condition that each data point will be a single line as shown below. It's good for the eyes, easy on git, easy to audit and easy to load into numpy. 

```
[
  [0.0, 62.4],
  [20.0, 63.1],
  [40.0, 63.5]
]
```

PR #482, opened 2026-04-01
Create GOVERNANCE.md

markussmiettinen:
As per request from @comcon1, and as discussed with @ohsOllila and @korbinib, I have created this minimal governance document.

----
📚 Documentation preview 📚: https://databank--482.org.readthedocs.build/


comcon1 (approved):

comcon1:
I've approved, but it would be nice to have also a minimal CoC which will refer this GOVERNANCE.md file. @korbinib I remember you offered to inherit some community-built CoC. Maybe we can also include it here?

PR #491, opened 2026-04-25
OP dataset generator (for use in Kaggle Dataset publishing)

comcon1:
- [x] Generate OP dataset from experiments.
- [x] Generate OP dataset from simulations.

The PR is a shorter version of the all-dataset version #476. It's much shorter for review.
This PR: https://github.com/NMRLipids/BilayerData/pull/388 already uses this branch.

It has additional dependencies: `tables` and `unidecode`, and it has very simple usage:
```sh
(Databank) comcon1@tanteliss:~/repo/Databank/developer$ ./gen-op-dataset.py --help
fairmd-lipids 1.5.1.dev4+g37cd713e8 by NMRlipids Open Collaboration - GPL-3.0-or-later
FAIRMD Lipids is initialized from the folder: /home/comcon1/repo/Databank/Data
---------------------------------------------------------------
usage: gen-op-dataset.py [-h] [--exps] [--sims]

options:
  -h, --help  show this help message and exit
  --exps      Run experiments
  --sims      Run simulations
```

It will generate one or another dataset in HDF5 format.

----
📚 Documentation preview 📚: https://databank--491.org.readthedocs.build/

### DIFF
--- a/developer/gen-op-dataset.py
+++ b/developer/gen-op-dataset.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+
+import numpy as np
+import pandas as pd
+from rdkit import Chem
+from rdkit.Chem import rdMolDescriptors
+from unidecode import unidecode
+
+from fairmd.lipids.experiment import ExperimentCollection, OPExperiment
+
+H5_MASTER = "exp-op-dataset.h5"
+
+
+def store_to_hdf5(df: pd.DataFrame, e: OPExperiment, lipid_name: str):
+    _mcontent = e.membrane_composition(basis="molar")
+    mcontent = {"name": [], "inchikey": [], "fraction": []}
+    for lname, frac in _mcontent.items():
+        k = lname
+        ik = e.lipids[lname].metadata["bioschema_properties"]["inChIKey"]
+        mcontent["name"] += [k]
+        mcontent["inchikey"] += [ik]
+        mcontent["fraction"] += [frac]
+    hydration = e.get_hydration()
+    scontent = e.solution_composition(basis="molar")
+    temperature = e["TEMPERATURE"]
+    inchikey = e.lipids[lipid_name].metadata["bioschema_properties"]["inChIKey"]
+    smiles = e.lipids[lipid_name].metadata["bioschema_properties"]["smiles"]
+    # store all vars and df to the HDF5 table
+    group = "E" + unidecode(e.exp_id).replace("-", "_").replace("/", "_").replace(".", "") + "__" + lipid_name
+
+    with pd.HDFStore(H5_MASTER, "a") as store:
+        # DataFrame table
+        store.put(f"{group}/op_values", df, format="table", data_columns=True)
+        # Metadata attributes
+        op_storer = store.get_storer(f"{group}/op_values")
+        op_storer.attrs["inchikey"] = inchikey
+        op_storer.attrs["smiles"] = smiles
+
+        # Metadata attributes
+        store.put(f"{group}/sample_table", pd.DataFrame(mcontent), format="table", data_columns=True)
+        sample_storer = store.get_storer(f"{group}/op_values")
+        sample_storer.attrs["temperature"] = temperature
+        sample_storer.attrs["hydration"] = hydration
+        sample_storer.attrs["solution"] = ", ".join([f"{k:<25} {v * 100:>6.1f}%" for k, v in sorted(scontent.items())])
+
+
+def by_c_uname(opdict: dict, uname: str) -> dict:
+    """Return only OP vals of named heavy atoms. Sorted by val."""
+    res = []
+    for k, v in opdict.items():
+        if k.split()[0] == uname:
+            res.append(v)
+    return sorted(res, key=lambda x: x[0])
+
+
+def main() -> None:
+    print("Generating OP datasets.")
+    ee = ExperimentCollection.load_from_data("OPExperiment")
+    for e in ee:
+        print(e)
+        for lname, opdict in e.data.items():
+            mol = e.lipids[lname]
+            smiles = mol.metadata["bioschema_properties"]["smiles"]
+            rdmol = Chem.MolFromSmiles("CCO")  # Ethanol
+            n_heavy_atoms = rdMolDescriptors.CalcNumHeavyAtoms(rdmol)
+            print(smiles)
+            smi2uname = {}
+            for uname, aprops in mol.mapping_dict.items():
+                if "SMILEIDX" in aprops:
+                    smid = int(aprops["SMILEIDX"])
+                    smi2uname[smid] = uname
+            if not smi2uname:
+                # NO SMILEIDX. Cannot store.
+                continue
+            smi2uname = dict(sorted(smi2uname.items()))
+            df = pd.DataFrame(
+                {
+                    "id": np.full(len(opdict), -1, dtype=np.int64),  # int64, init NaN
+                    "val": np.full(len(opdict), np.nan),
+                    "err": np.full(len(opdict), np.nan),
+                }
+            )
+            cur_row = 0
+            for id_, uname in smi2uname.items():
+                for opdict_row in by_c_uname(opdict, uname):
+                    if cur_row >= len(df):
+                        break
+                    df.at[cur_row, "id"] = id_
+                    df.at[cur_row, "val"] = opdict_row[0]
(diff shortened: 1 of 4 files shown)


developer/gen-op-dataset.py
```diff
@@ -0,0 +1,252 @@
...
+            }
+        )
+        cur_row = 0
+        for id_, uname in smi2uname.items():
+            for opdict_row in OPDataStorer.by_c_uname(opdict, uname):
+                if cur_row >= len(df):
+                    break
+                df.at[cur_row, "id"] = id_
+                df.at[cur_row, "val"] = opdict_row[0]
+                df.at[cur_row, "err"] = err_extractor(opdict_row)
```
batukav: The dataframe is preallocated with `len(opdict)` rows, but `cur_row` only advances for atoms that have a `SMILEIDX` mapping. Whenever some entries in `opdict` don't map (and that's a legitimate case — that's the whole reason `smi2uname` filters on `SMILEIDX`), the trailing rows stay at their init values `(id=-1, val=NaN, err=NaN)` and get written into the HDF5 store as garbage rows.

A couple of options:
- Truncate at the end: `self._df = df.iloc[:cur_row].reset_index(drop=True)`.
- Or build a list of dicts/tuples and construct the DataFrame from it once at the end — same shape, no preallocation guesswork.

Also, the `if cur_row >= len(df): break` guard hides the opposite mismatch (more matched rows than `len(opdict)` — shouldn't happen but if it ever does, you'll silently truncate). I'd remove the guard once you switch to one of the patterns above.

developer/gen-op-dataset.py
```diff
@@ -0,0 +1,252 @@
...
+    def prepare_dataframe(self):
+        """Prepare dataframe for storing"""
+
+    @property
+    @abstractmethod
+    def ass_id(self):
+        """Get id of assoc object"""
+
+    def _prepare_df_common(self, mol, opdict, err_extractor):
+        """Common code for dataframe conditioning for storage"""
```
batukav: Looks like a stray debug `print(smiles)` — it'll fire once per lipid per system. Worth removing or moving behind a logger.

developer/gen-op-dataset.py
```diff
@@ -0,0 +1,252 @@
...
+        opdict = self._e.data[self._lname]
+        self._prepare_df_common(mol, opdict, err_extractor=lambda x: 0.02 if len(x) == 1 else x[1])
+
+
+class SimOPDataStorer(OPDataStorer):
+    @property
+    def ass_id(self):
+        return self._s["ID"]
+
+    def prepare_dataframe(self):
```
batukav, line 180: `store_to_hdf5` in `ExpOPDataStorer` and `SimOPDataStorer` are ~90% identical: same DataFrame `put` calls, same metadata loop, same temperature/hydration/solution attribute block. Only the group-name construction, the membrane-content schema, and the per-storer-specific attrs (`fmdl_expid`/`nmr_method` vs `fmdl_simid`/`ff_name`) really differ.

Since you already have an abstract base, this is a natural place to push the shared scaffolding up — e.g. abstract methods like `_group_name()`, `_membrane_content()`, `_op_attrs()`, `_sample_attrs()`, with one concrete `store_to_hdf5` on the base. Would shrink the file noticeably and prevent the two paths from drifting.

developer/gen-op-dataset.py
```diff
@@ -0,0 +1,252 @@
...
+        scontent = self._e.solution_composition(basis="molar")
+        temperature = self._e["TEMPERATURE"]
+        inchikey = self._e.lipids[self._lname].metadata["bioschema_properties"]["inChIKey"]
+        smiles = self._e.lipids[self._lname].metadata["bioschema_properties"]["smiles"]
+        nmr_method = self._e.metadata.get("NMR", {}).get("METHOD", False)
+        # store all vars and df to the HDF5 table
+        group = "E"
+        group += (
+            unidecode(self._e.exp_id)
+            .replace("-", "_")  # prefer variable-like strs
```
batukav: The chained `.replace("-", "_").replace("/", "_").replace(".", "")` is brittle — any new character that turns up in an `exp_id` (spaces, colons, parentheses, `+`, etc.) silently produces an invalid HDF5 group name. A single regex like `re.sub(r"[^A-Za-z0-9_]", "_", unidecode(self._e.exp_id))` covers the general case in one line.

developer/gen-op-dataset.py
```diff
@@ -0,0 +1,252 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--exps", action="store_true", help="Run experiments")
+    parser.add_argument("--sims", action="store_true", help="Run simulations")
+    args = parser.parse_args()
+
+    if args.exps:
+        main_exps()
+    if args.sims:
+        main_sims()
```
batukav: If the user runs the script with neither `--exps` nor `--sims`, it succeeds silently and produces nothing. Worth either printing `parser.print_help()` and exiting non-zero, or making it a mutually-exclusive required group, so the no-flag case fails loudly.

developer/gen-op-dataset.py
```diff
@@ -0,0 +1,252 @@
...
+            # DataFrame table
+            store.put(f"{group}/op_values", self._df, format="table", data_columns=True)
+            store.put(f"{group}/sample_table", pd.DataFrame(mcontent), format="table", data_columns=True)
+            # Metadata attributes - I
+            op_storer = store.get_storer(f"{group}/op_values")
+            upd_attr = {
+                "inchikey": inchikey,
+                "smiles": smiles,
+                "fmdl_expid": self._e.exp_id,
+            }
```
batukav: Nit: the magic `0.02` default error appears here and again in #495. Worth promoting to a module-level constant (e.g. `DEFAULT_OP_ERR = 0.02`) so it's documented in one place.

developer/gen-op-dataset.py
```diff
@@ -0,0 +1,252 @@
...
+        print(sim)
+        opdata = get_OP(sim)
+        if opdata is None:
+            continue
+        for lname in opdata:
+            if opdata[lname] is None:
+                continue
+            ods = SimOPDataStorer(sim, lname)
+            try:
+                ods.prepare_dataframe()
```
batukav, line 254: Both `main_exps` and `main_sims` swallow `OPDataError` with a printed warning and continue. For a Kaggle-published dataset it would be useful to know up front how many systems were skipped and why — at minimum a final summary line like `"Skipped N of M systems"`, ideally with skip reasons. Otherwise it's hard to tell whether the resulting HDF5 covers what you expect.

batukav (commented):
Might be worth fixing before merge is the dataframe preallocation in `_prepare_df_common`, which can leave NaN/`-1` rows in the output when not every `opdict` entry has a `SMILEIDX` mapping. Beyond that there's a fair amount of duplication between the two `store_to_hdf5` implementations that the abstract base could absorb. Minor cleanups noted inline.

developer/gen-op-dataset.py
comcon1: removed

developer/gen-op-dataset.py
comcon1: I made a cycle that just appends the dataframe. It's indeed better.

developer/gen-op-dataset.py
comcon1: That's good. Replaced.

developer/gen-op-dataset.py
comcon1: Done here: 89c3f6d6

developer/gen-op-dataset.py
comcon1: FIxed in d335cc920970717249cabca53ed8e21f7b008345

developer/gen-op-dataset.py
comcon1: Logger is plugged in. Statistics is counted now also. Done in 7df8d8046b02d292e7b8863310cbbf8a13eeb233

developer/gen-op-dataset.py
comcon1: will take some time. I've done all except this one. Still working on it..

comcon1: I did quite a lot of method extractions and moving up to base class: 90f4b3748a8786b734152cbfdafab9ddd258f7e8

batukav (approved):

PR #495, opened 2026-05-10
Add atomic point quality dataset generator

comcon1:
This script generates the point quality dataset of the following schema:

| c | h | val_s | err_s  | val_e | err_e |
|--|--|--|--|--|--|
| Carbon SMILE ID | H ID | OP value (simulatioon) | err | OP value (experiment), with the sign from the simulation |err |

For a single carbon atom, if proton's OP value cannot be different (there is only one proton or 3 protons), then H ID is only "1" and only single value is used. 

For CH2 groups, protons can be different, we assign 1 to be proton with smaller absolute OP and 2 with larger.
This clears the problem of the uncertainty of chirality that we have for some atoms.

We do not trust the sign of the OP value in any experiments. Signs are always inherited from the simulation.

All values from all experiments are averaged using only `c,h` pair as a key.

----
📚 Documentation preview 📚: https://databank--495.org.readthedocs.build/

### DIFF
--- a/developer/gen-atomicq-dataset.py
+++ b/developer/gen-atomicq-dataset.py
@@ -13,6 +13,10 @@ class OPQDataError(Exception):
 
 
 class OPQDataStorer():
+
+
+    DEFAULT_OPQ_HDFNAME = "atomic-opq-dataset.h5"
+
     def __init__(self, s: System, lname: str) -> None:
         self._s = s
         self._lname = lname
@@ -143,7 +147,6 @@ class OPQDataStorer():
                 sample_storer.attrs[k] = v
 
 
-H5_QSIMS_MASTER = "atomic-opq-dataset.h5"
 
 
 def load_sims() -> None:
@@ -170,7 +173,7 @@ def load_sims() -> None:
             ods.average_experiment_data()
 
             ods.compute_q_points()
-            ods.store_to_hdf5(H5_QSIMS_MASTER)
+            ods.store_to_hdf5(OPQDataStorer.DEFAULT_OPQ_HDFNAME)
 
 
 if __name__ == "__main__":
--- a/developer/gen-atomicq-dataset.py
+++ b/developer/gen-atomicq-dataset.py
@@ -5,7 +5,7 @@ import pandas as pd
 
 from fairmd.lipids.api import get_OP
 from fairmd.lipids.core import System, initialize_databank
-from fairmd.lipids.experiment import ExperimentCollection
+from fairmd.lipids.experiment import ExperimentCollection, OPExperiment
 
 
 class OPQDataError(Exception):
@@ -16,7 +16,7 @@ class OPQDataStorer():
     def __init__(self, s: System, lname: str) -> None:
         self._s = s
         self._lname = lname
-        self._exp_opdicts = []
+        self._exp_opdicts: list[pd.DataFrame] = []
 
     def _cvt_op_df(self, opdict: dict, err_pos: int) -> pd.DataFrame:
         op_clean = pd.DataFrame(
@@ -25,7 +25,7 @@ class OPQDataStorer():
         smi2uname = self._get_smi2uname()
         for smid, uname in smi2uname.items():
             _c_dict = {
-                k: [v[0], 0.02 if err_pos >= len(v) else v[err_pos]]  # check
+                k: [v[0], OPExperiment.DEFAULT_ERROR if err_pos >= len(v) else v[err_pos]]  # check
                 for k, v in opdict.items()
                 if k.split()[0] == uname
             }
@@ -42,7 +42,11 @@ class OPQDataStorer():
                 op_clean.loc[len(op_clean)] = [smid, 1, vearr[0, 0], vearr[0, 1]]
                 op_clean.loc[len(op_clean)] = [smid, 2, vearr[1, 0], vearr[1, 1]]
             else:
-                msg = f"Unexpected number of H for {uname} in instance {self.ass_id} // {self._lname}: {_c_dict_len}. Cannot store."
+                msg = (
+                    f"Unexpected number of H for {uname} in "
+                    f"instance {self.ass_id} // {self._lname}: {_c_dict_len}."
+                    " Cannot store."
+                )
                 raise OPQDataError(msg)
         return op_clean.astype({"c": np.int64, "h": np.int64, "val": np.float64, "err": np.float64})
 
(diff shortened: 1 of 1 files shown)


developer/gen-atomicq-dataset.py
```diff
@@ -0,0 +1,177 @@
...
+            # NO SMILEIDX. Cannot store.
+            msg = f"Instance {self.ass_id} // {self._lname} cannot be stored: we don't have SMILEIDX."
+            raise OPQDataError(msg)
+        return dict(sorted(s2u.items()))
+
+    def add_experiment_data(self, exp_opdict: dict) -> None:
+        """Add experimental OP data to the storer. We will use it for Q estimation."""
+        if not hasattr(self, "_exp_opdict"):
+            self._exp_opdicts = []
+        self._exp_opdicts.append(self._cvt_op_df(exp_opdict, 1))
```
batukav: Looks like a bug: the `hasattr` check is for `_exp_opdict` (singular) but the attribute we actually set/append to is `_exp_opdicts` (plural). Since `_exp_opdict` is only ever set later in `average_experiment_data`, this guard is always falsy when `add_experiment_data` runs, so `self._exp_opdicts = []` resets the list on every call and only the *last* experiment ends up being averaged.

In the loop in `load_sims` you iterate over `paired_opedict[lname]` and call `add_experiment_data` once per experiment, so for any lipid with more than one experimental record the averaging silently drops everything but the last one.

Easiest fix is to either match the name (`hasattr(self, "_exp_opdicts")`) or just initialize `self._exp_opdicts = []` in `__init__`.

developer/gen-atomicq-dataset.py
```diff
@@ -0,0 +1,177 @@
...
+            }
+            _c_dict_len = len(_c_dict)
+            if _c_dict_len == 0:
+                continue
+            vearr = np.array([_c_dict.popitem()[1] for _ in range(_c_dict_len)])
+            if _c_dict_len == 1:  # one H
+                op_clean.loc[len(op_clean)] = [smid, 1, vearr[0, 0], vearr[0, 1]]
+            elif _c_dict_len == 3:  # three H. They are always symmetric.
+                op_clean.loc[len(op_clean)] = [smid, 1, np.mean(vearr[:, 0]), np.mean(vearr[:, 1])]
+            elif _c_dict_len == 2:  # four H. They could be asymmetric.
```
batukav: Comment says "four H" but this branch is the CH2 case (two protons) — the PR description even spells this out. Worth fixing the comment so future readers don't get confused.

developer/gen-atomicq-dataset.py
```diff
@@ -0,0 +1,177 @@
...
+                vearr = vearr[np.argsort(np.abs(vearr[:, 0]))]
+                op_clean.loc[len(op_clean)] = [smid, 1, vearr[0, 0], np.mean(vearr[0, 1])]
+                op_clean.loc[len(op_clean)] = [smid, 2, vearr[1, 0], np.mean(vearr[1, 1])]
```
batukav: `vearr[0, 1]` and `vearr[1, 1]` are scalars at this point, so `np.mean(...)` is a no-op. This looks like a copy-paste leftover from the symmetric three-H branch above. If you actually meant to combine the two error values (which would defeat the asymmetric split), the current code doesn't do that.

developer/gen-atomicq-dataset.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @property
+    def ass_id(self) -> str:
+        """Get id of assoc object"""
+        return self._s["ID"]
+
+    def _get_smi2uname(self) -> dict:
+        mol = self._s.lipids[self._lname]
+        smiles = mol.metadata["bioschema_properties"]["smiles"]
+        print(smiles)
```
batukav: Stray `print(smiles)` looks like leftover debugging — it'll fire on every lipid for every system processed. Either remove it or route it through a logger so the noise can be silenced.

developer/gen-atomicq-dataset.py
```diff
@@ -0,0 +1,177 @@
...
+from fairmd.lipids.api import get_OP
+from fairmd.lipids.core import System, initialize_databank
+from fairmd.lipids.experiment import ExperimentCollection
+
+
+class OPQDataError(Exception):
+    """Our specific exception"""
+
+
+class OPQDataStorer(ABC):
```
batukav: Inheriting from `ABC` here is misleading: there are no `@abstractmethod`s, so the class is fully instantiable and behaves like a regular class. Either drop the ABC inheritance or, if you intended to share structure with `OPDataStorer` from #491, lift the common bits into a real shared base.

developer/gen-atomicq-dataset.py
```diff
@@ -0,0 +1,177 @@
...
+        for smid, uname in smi2uname.items():
+            _c_dict = {
+                k: [v[0], 0.02 if err_pos >= len(v) else v[err_pos]]  # check
+                for k, v in opdict.items()
+                if k.split()[0] == uname
+            }
+            _c_dict_len = len(_c_dict)
+            if _c_dict_len == 0:
+                continue
+            vearr = np.array([_c_dict.popitem()[1] for _ in range(_c_dict_len)])
```
batukav: Minor: building `vearr` by repeatedly calling `popitem()` in a comprehension is harder to read than `np.array(list(_c_dict.values()))` and depends on dict insertion order. Same result, less ceremony.

developer/gen-atomicq-dataset.py
```diff
@@ -0,0 +1,177 @@
...
+        self._lname = lname
+
+    def _cvt_op_df(self, opdict: dict, err_pos: int) -> pd.DataFrame:
+        op_clean = pd.DataFrame(
+            columns=["c", "h", "val", "err"],
+        )
+        smi2uname = self._get_smi2uname()
+        for smid, uname in smi2uname.items():
+            _c_dict = {
+                k: [v[0], 0.02 if err_pos >= len(v) else v[err_pos]]  # check
```
batukav: The `0.02` default error is also hardcoded in #491 with the same magic number. Worth pulling it out into a module-level constant (e.g. `DEFAULT_OP_ERR = 0.02`) so it's easy to find and document the assumption.

batukav (commented):
Left a few inline comments. The most important one is the `hasattr` typo in `add_experiment_data` that resets `_exp_opdicts` on every call — as written, lipids with multiple paired experiments will only have the last one survive into the average. The rest are smaller cleanups.

developer/gen-atomicq-dataset.py
comcon1: Oh. A big mistake.

comcon1: done: 537003ffa659729c8f63c73b1cf0b4d58323d65b

developer/gen-atomicq-dataset.py
comcon1: done

developer/gen-atomicq-dataset.py
comcon1: no, it's just a copy-paste. np.mean removed now

developer/gen-atomicq-dataset.py
comcon1: removed

developer/gen-atomicq-dataset.py
comcon1: It's a bit hard to merge them. Let's just remove ABC

developer/gen-atomicq-dataset.py
comcon1: thx. replaced.

developer/gen-atomicq-dataset.py
comcon1: let's merge #491 first, then this constant will come to the API. And then we will use it here. Will set as draft before that.

comcon1: done

batukav (approved):

PR #504, opened 2026-06-07
Proper computation of SYSTEM quality

comcon1:
According to documentation https://databank.readthedocs.io/stable/quality.html
system quality should be multiplied by (1-w) for *each fragment*. In the code, we had (1-w) computed for whole lipids for all fragments and then П(1-w_i) was multiplied to all averaged fragment qualities. It was a bug.

Now, the situation is fixed. We check fragment-by-fragment. If the headgroup quality is nan for 1 lipid, then we do `penalty["head"] *= (1-w)` where `w` is a molar fraction. It now doesn't affect at all the quality of the tails, which is proper. Now, it looks proper. Database should be recomputed after this merge is done.

It's a small but very important fix.

Fixed qualities are visible in this branch https://github.com/NMRLipids/BilayerData/commits/403-view-fixed-qualities/. One may pay attention to this commit: https://github.com/NMRLipids/BilayerData/commit/778dda9821a7f77094819f5e5daa832216305901

----
📚 Documentation preview 📚: https://databank--504.org.readthedocs.build/

### DIFF
--- a/src/fairmd/lipids/quality.py
+++ b/src/fairmd/lipids/quality.py
@@ -327,13 +327,13 @@ class OPQualityEvaluator(QualityEvaluator):
                     lip_tails += value  # everything non head is tail??
             w = molar_composition[lname]
             if lip_total == 0:
-                penalty["total"] *= (1-w)
+                penalty["total"] *= 1 - w
             total += lip_total
             if lip_head == 0:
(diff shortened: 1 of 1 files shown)


markussmiettinen (approved):
Looks good to me.

PR #510, opened 2026-06-14
Fix tests-develop test

### LINKED ISSUE #508: CI tests of development scripts don't work
**Describe the bug**
The bug is visible here
https://github.com/NMRLipids/FAIRMD_lipids/actions/runs/27496306343/job/81270887384

**To Reproduce**
Just run Test development scripts.

**Expected behavior**
Data initialization with ToyDatabank should be enough. Unit tests work with it and dev tests should also do the same.

**Actual behavior**
Data is

mdondrup:
This PR fixes regression #508 introduced in #506 by loading fairmd.lipids module during a test stage which the test wasn't designed for. 

When imported the fairmd.lipids package does some heavy lifting, whose __init__.py raises RuntimeError when FMDL_DATA_PATH points to a non-existent folder. Develop tests shouldn't trigger that heavy package initialization, so loading the schema from the filesystem keeps them isolated.

Now loading the schema directly again. If you think, loading the package initialization is necessary, you need to provide it with the proper environment in the test.

(Btw, did I mention that unconditional loading of data from an environment variable was a bad idea? __init__ should be lazy loading and side-effect free ) 

----
📚 Documentation preview 📚: https://databank--510.org.readthedocs.build/

### DIFF
--- a/tests/develop/test_autocomplete_metadata.py
+++ b/tests/develop/test_autocomplete_metadata.py
@@ -1,4 +1,3 @@
-from importlib.resources import files
 import importlib.util
 import json
 import os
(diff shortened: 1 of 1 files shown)


mdondrup:
```
tox -e tests-develop
[...]

tests/develop/test_autocomplete_metadata.py ..                           [100%]

====================== 2 passed, 205 deselected in 6.07s =======================
  tests-develop: OK (12.66=setup[5.56]+cmd[0.09,7.00] seconds)
  congratulations :) (13.68 seconds)
```

comcon1:
I will not approve it. If we use something from the package, we must import it from the package. The error should be fixed in the manifests.

mdondrup:
> I will not approve it. If we use something from the package, we must import it from the package. The error should be fixed in the manifests.
 
 Ok just fix it the way you want.

PR #511, opened 2026-06-14
Fix dev-tests and make them following the same mechanism of importing as unit-tests

comcon1:
This scheme allows develop-scripts and their tests to import from fairmd.lipids data using ToyData as a data folder. It's not very important for the current particular test initiated by @mdondrup; however, it's of highly importance for future testing of other scripts in `develop` because we allow (and recommend) importing `fairmd.lipids` from develop scripts.

----
📚 Documentation preview 📚: https://databank--511.org.readthedocs.build/

### DIFF
--- a/.github/workflows/tests-develop.yml
+++ b/.github/workflows/tests-develop.yml
@@ -6,6 +6,7 @@ on:
     paths:
       - "developer/**"
       - "tests/develop/**"
+      - "tests/conftest.py"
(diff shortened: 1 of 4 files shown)


comcon1:
Now, we can see that testing developmental scripts was triggered and passed successfully.

mdondrup (approved):
Great. Test pass.

PR #513, opened 2026-06-22
retry on transient failures

korbinib:
- retry on transient failures (HTTP 429/500/502/503/504 and connection-level errors like timeouts/dropped connections), but do not retry definitive errors like 404.
- Polite backoff
- User agent
- AUTOCOMPLETE_MAX_RETRIES env var (default 4)

trade-off: if one of the API endpoints is entirely down there is now ~15s retrying before giving up, vs. failing fast before - considered acceptable, since we run this asynchronously 

----
📚 Documentation preview 📚: https://databank--513.org.readthedocs.build/

### DIFF
--- a/developer/autocomplete_metadata.py
+++ b/developer/autocomplete_metadata.py
@@ -12,36 +12,105 @@ queried with the inchikey from:
 
 .. note::
    This file is meant to be used by automated workflows.
+
+   Several upstream services (notably EBI's UniChem and ChEBI) intermittently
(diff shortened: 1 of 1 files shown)


mdondrup (approved):
This looks like a robust improvement. I think we may need to reconsider timeouts and retries in case it doesn't work. On the other hand, I think it isn't necessary to overthink it and adapt and fine-tune our service to compensate for the deficiencies of another.

PR #523, opened 2026-09-09
Bioschema dataset profile

korbinib:
FAIRMD_lipids part of https://github.com/NMRLipids/BilayerData/issues/490

----
📚 Documentation preview 📚: https://databank--523.org.readthedocs.build/


docs/src/schemas/experiment_metadata.md
```diff
@@ -113,9 +122,93 @@ For NMR sample, it is important to mention how the targeted hydration level is r
...
+bioschema_properties:
+  name: Fluid phase lipid areas and bilayer thicknesses of commonly used phosphatidylcholines
+  description: Experimental X-ray scattering form factor for a lipid bilayer containing POPC ...
+  sameAs: https://doi.org/10.1016/j.bbamem.2011.07.022
+  datePublished: '2011-11'
+  license:
+    spdx: CC-BY-4.0
+    name: Creative Commons Attribution 4.0 International
+    url: https://spdx.org/licenses/CC-BY-4.0.html
+    sameAs: https://creativecommons.org/licenses/by/4.0/
```
comcon1, line 202: Where did this `license` come from? Can you write a small clarification about this `license` and that it is not `articleLicense` and why?

docs/src/schemas/simulation_metadata.md
```diff
@@ -205,3 +208,80 @@ When adding a simulation into FAIRMD Lipids with the [fmdl_add_simulation](add_s
...
+  alternateName: 200POPC_9000SOL_313K
+  description: POPC CHARMM36 T313K. The simulations are ran for ...
+  sameAs: https://doi.org/10.5281/zenodo.4040423
+  datePublished: '2020-09-21'
+  license:
+    spdx: CC-BY-4.0
+    name: Creative Commons Attribution 4.0 International
+    url: https://spdx.org/licenses/CC-BY-4.0.html
+    sameAs: https://creativecommons.org/licenses/by/4.0/legalcode
+  publisher: Zenodo
```
comcon1, line 267: Here I understand that license comes from Zenodo terms, right? But it is also woth clarifying. For cases, where not only zenodo is used.

comcon1 (changes requested):
Thanks for this workflows! I requested a couple of clarifications. Probably you discussed it in a workshop, but I think it's good to add.

As it is a workflow, it will be easier to test natively than try to find issues during review. I will try to generate zoo/luna AI review and will see if it's worth publishing it.

docs/src/schemas/simulation_metadata.md
korbinib: The workflow will query the datacite and the crossref API and take the respective licenses for full datasets.
So since we are using the datacite and not the zenodo, this also works for dataverse etc.

docs/src/schemas/experiment_metadata.md
korbinib: We have discussed this in length at the workshop. The data itself is a fact and is not creative work, hence it does not reach the threshold of originality.
The IPR with the higher relevance here is the database directive, but the effort is created from back-digitalising plots, hence the license of the repo (FAIRMD) should apply (expect in cases where the data is provided already as a dataset itself)

developer/autocomplete_expsim_metadata.py
```diff
@@ -1160,7 +1222,7 @@ def _listed(entries):
     return ", ".join(entries[:-1]) + " and " + entries[-1]
 
 
-def compose_description(readme, path, kind, block, names):
+def compose_description(readme, path, kind, block, dois, names):
```
comcon1: I think, this function is very important. Let's refactor it. Can you please extract functions defining description for all three types of analyses. Maybe to a different file + import it.

developer/autocomplete_expsim_metadata.py
comcon1: Let's refactor this: move constants to a separated import-file.

comcon1: Up to here - it's helper functions. They can be separated.

comcon1, line 94: 1700 Lines until main for CLI!

comcon1 (changes requested):
Thanks for previous changes. Now we can pay attention to breaking appart the autocomplete script. It's really a mastodon script! It's actually written quite nicely. Maybe a bit separation + maybe some incapsulation. Even LLM-assisted could be fine refactoring for it.

docs/src/schemas/simulation_metadata.md
comcon1: I see. Thanks for the clarifications!

korbinib:
@comcon1 I have split this now in multiple helpers and separated out the constants.

developer/expsim_metadata/records.py
```diff
@@ -0,0 +1,200 @@
...
+    hand-written reference cannot be lost to a parsing slip.
+    """
+    doc = yaml.safe_load(text) or {}
+    if "PUBLICATION" not in doc:
+        return text, False
+    cites = [str(c) for c in (block.get("citation") or [])]
+    for item in parse_publication(doc.get("PUBLICATION")):
+        if item not in cites:
+            print(f"  keeping PUBLICATION: {item!r} is not in citation")
+            return text, False
```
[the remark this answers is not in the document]
korbinib: will be addressed when the legacy data is gone

comcon1 (approved):
This is very solid work and a large PR (I currently try to work in a smaller PRs, this one is a bit out of human-readable size). I approve it. Here are some comments for future development:
1. We agreed to use ruff and we have ruff-dev.toml that lists the exceptions. We don't need to strictly satisfy ruff but it's good at least trying to satisfy it.
2. Current refactoring made the code much more readable. There is still place for refactoring that can be done later. Especially considering OOP as we have our API hierarchy for experiment and simulation objects. Exps are to type and there is also an abstract class that owns the composition. Some functions can be considered to be moved to API.
3. I don't believe that someone will review the schema so maybe hanging all this on the autocomplete is not the best way. But we can try and see how it works after a year for example. Current flow can be adopted in a way that bioschema properties are constructed during DB population.

Issue #512, opened 2026-06-20
Restrictions on redundand atom names


comcon1:
**Do you know if it is related to any opened issues?**
It's related to an issue in BilayerData: https://github.com/NMRLipids/BilayerData/issues/427

**Where it should appear**
- [ ] Adding new molecule https://github.com/NMRLipids/BilayerData/issues/427
- [ ] On the page mols&mapping https://databank.readthedocs.io/stable/schemas/moleculesAndMapping.html

PR #525, opened 2026-09-19
Fix information in docs (info files and atom uniqueness)

### LINKED ISSUE #423: Remove INFO-file description from Data structure
Info-files are removed and will not be stored in BilayerData anymore.
https://nmrlipids.github.io/FAIRMD_lipids/stable/dbstructure.html#other-files
It should be written instead of description of info-files

comcon1:
- Fix information in docs that we don't store info files anymore. Info-files - are now just an input to create README automatically.
- Add notes about obligatory atom uniqueness

The task is partly done by Zoo/Luna with my curation.

It will close #423 #512 

Documentation changes are visible here: 
- https://databank--525.org.readthedocs.build/525/contrib/addingMolecule.html#mapping-file-creation
- https://databank--525.org.readthedocs.build/525/contrib/addingSimulation.html
- https://databank--525.org.readthedocs.build/525/dbstructure.html#


comcon1:
Testing developmental scripts fails due to workflow setting problems (I made issue #526). Here it's not relevant - we can ignore it.

docs/src/contrib/addingMolecule.rst
```diff
@@ -74,10 +73,15 @@ add the residue name to ``RESIDUE`` key of each atom in the mapping file. In thi
...
-different force fields, you can use more general naming.
+The mapping file should contain all the atoms of the molecules and should be named
+``mapping-YOURMOL-usecase.yaml``. Very often, the usecase is the force field name, however, if the
+naming convention is the same between different force fields, you can use more general naming.
+
+.. warning::
+
+   Every atom must be uniquely identifiable by its simulation-specific ``ATOMNAME`` and ``RESIDUE``
+   combination. This requirement is especially important for mappings used with lipid force fields
+   that split a molecule into several residues. If you are adding a new molecule, and it has name collisions, we don't accept this molecule and this simulation into the databank. You will need to change the names and resubmit.
```
hsantila: It is maybe not evident what is meant by name collision? Does it only refer to the case where some atoms are not uniquely identifiable?

hsantila (changes requested):
Naming collision was an unclear term for my morning brain :)

docs/src/contrib/addingMolecule.rst
comcon1: I rephrased. You can also add suggestions right in the comments if you want.

PR #531, opened 2026-09-21
workaround for EOL issue

### LINKED ISSUE #536: PR suggestions from AutocompleteExpSimMetadata do not get populated
The anchor for reviewdog is EOL of the last line with the length of the suggestions we add via  AutocompleteExpSimMetadata.yml this fails and the suggestion is not added, while inline suggestions earlier are.

korbinib:
together with https://github.com/NMRLipids/BilayerData/pull/537 closes https://github.com/NMRLipids/BilayerData/issues/536 

----
📚 Documentation preview 📚: https://databank--531.org.readthedocs.build/


comcon1 (commented):
Should we remove PUBLICATION from ToyData as well?

comcon1:
@korbinib, I found that the package started to accumulate a large tech debt upon the addition of metadata handlers, so I added some refactorings here above your commits. Hope they make the codebase cleaner (at least a bit)..

comcon1 (approved):
I will unblock now. You can verify my refactoring and merge.

korbinib:
> Should we remove PUBLICATION from ToyData as well?

Yes, I think I also should update the ToyData README.yml

